PR #96, opened 2025-06-23
feat: add Namestone Integration for Profile Management

lourou:
Automatically registers and manages usernames with Namestone when users create or update their profiles.

## Changes
- **New**: `src/utils/namestone.ts` - Namestone service with `setName()` and `deleteName()` methods
- **Modified**: `src/api/v1/users/handlers/create-user.ts` - Auto-register username on profile creation
- **Modified**: `src/api/v1/profiles/handlers/update-profile.ts` - Handle username changes and metadata updates
- **Updated**: `.env.example` - Added Namestone config variables

## Configuration
```env
NAMESTONE_BASE_URL=
NAMESTONE_API_KEY=
NAMESTONE_DOMAIN=convosapp.eth | convos.org
```

## Behavior
- **Profile creation**: Registers username with Namestone (async, non-blocking)
- **Profile updates**: 
  - Username change → delete old name, set new name
  - Metadata change → update text records
- **Error handling**: Namestone failures are logged but don't block operations
- **Graceful degradation**: Works without Namestone configuration

All Namestone operations are asynchronous and won't affect user experience if they fail

## Summary by CodeRabbit

- **New Features**
  - Integrated Namestone service to automatically register and update usernames and related profile information with the Namestone API when users are created or profiles are updated.
- **Chores**
  - Updated environment variable examples to include new Namestone configuration options.

### DIFF
--- a/src/utils/namestone.ts
+++ b/src/utils/namestone.ts
@@ -28,13 +28,18 @@ class NamestoneService {
   constructor() {
     this.apiKey = process.env.NAMESTONE_API_KEY || "";
     this.domain = process.env.NAMESTONE_DOMAIN || "";
-    this.baseUrl = process.env.NAMESTONE_BASE_URL || "https://namestone.com/api/public_v1";
+    this.baseUrl =
+      process.env.NAMESTONE_BASE_URL || "https://namestone.com/api/public_v1";
 
     if (!this.apiKey) {
-      console.warn("NAMESTONE_API_KEY not configured - Namestone integration disabled");
+      console.warn(
+        "NAMESTONE_API_KEY not configured - Namestone integration disabled",
+      );
     }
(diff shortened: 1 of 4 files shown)


src/utils/namestone.ts
```diff
@@ -0,0 +1,165 @@
...
+    return !!(this.apiKey && this.domain);
+  }
+
+  /**
+   * Sets a name (subdomain) for a given address and domain
+   */
+  async setName(args: {
+    username: string;
+    address: string;
+    textRecords?: Record<string, string>;
```
thierryskoda, line 56: Maybe we can type this better. Like the textRecords, can they key be anything? or only some specific ones?

thierryskoda (approved):
Small typing improvement but otherwise looks good to me!

PR #117, opened 2025-08-13
Make docker compose use the local .env file if it exists

nmalzieu:
Tests that were relying on the notifications go server to notify our backend via webhook were failing
Fix in two parts
- this PR forces docker-compose to use the local .env file if it exists so the webhook header secret will be well set
- when running tests you need to have the env variable `TEST_NOTIFICATION_DELIVERY_URL` set to `http://host.docker.internal:8081/` . This is currently done in the `test:notifications-client` command that Thierry had created - this means that running only `bun test` will still result in some the notifications-client tests to fail but I guess that was the case before - do they run well on the CI @lourou or does this need fixing? 

## Summary by CodeRabbit

* **Chores**
  * Dev environment now automatically uses a local .env file when present for Docker Compose runs.
* **Tests**
  * Corrected test script to use the proper Bun command for running notification client tests.
  * Removed unnecessary console output from a notification test for cleaner logs.

### DIFF
--- a/dev/compose
+++ b/dev/compose
@@ -1,4 +1,9 @@
 #!/bin/bash
 set -eou pipefail
 
-docker compose -f dev/compose.yml -p "convos" "$@"
+# Pass the .env file if it exists locally - useful to run tests locally
+if [ -f .env ]; then
+  docker compose --env-file .env -f dev/compose.yml -p "convos" "$@"
+else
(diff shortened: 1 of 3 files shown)


lourou (approved):

lourou:
Yes @nmalzieu push notifications server tests run and pass on CI

PR #141, opened 2025-10-09
Backend v2 foundations

lourou:
### Introduce v2 backend foundations by adding v2 auth, device registration, notifications routing, and database models for push with unique APNS token constraint and MAX_PUSH_FAILURES=10
- Add v2 database models and constraints for push notifications with `DeviceRegistration` and `ClientIdentifier`, including a unique index on `(pushTokenType, apnsEnv, pushToken)` and nullable `pushToken` in [schema.prisma](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-5b443964f4f3a611682db8f7e02177b0a8c632b2039e2bd5e4dd7347815c565c) with corresponding migrations [20251003172324_add_v2_push_notification_models](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-ebbab47c87ae3238b65960a64b6598233c447afa532db36a15efcd73b3a9a7f2) and [20251017193536_add_unique_constraint_to_push_tokens](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-f9468b4aeccb07ff2c1e98fc16aef0a1e3e63e2fc0ef9d5af87f5d9ee1df1775).
- Add shared XMTP webhook handler and router to support v2 and v1 flows, validation, APNS send, failure tracking with `MAX_PUSH_FAILURES=10`, device disablement, and cleanup in [webhook-handler.ts](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-9ee12a996c2234bd60a6517d734ac22eb96ef397eb6f6b763710d4d8047df9c7) and [webhook.router.ts](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-debd7a4178638aad1420d02ea2eac3954c9b3e44f06c80757dc01a401bb18011), and mount in v1 and v2 index routers.
- Add v2 routes for auth, device, and notifications, including `/auth/token` with rate limit 100/5m and AppCheck-only, `/device/register` with APNS-only token handling and uniqueness reassignment, and notifications subscribe/unsubscribe/unregister in [v2 index](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-2027b6d52fe007c8b0c1dd4b9196c1d432b42462c0efaa178004eca83ef794eb), [auth.router.ts](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-a64941e2e2a710a168f51e763a59c5eeca14f1145390b7da30ef7882d763b9c0), [generate-token.ts](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-efae9276d245e77d478e0961baa0a511ec789b80927164138b3523f31781fb70), [device.router.ts](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-e12e88b9325b3fedf66bbd1062fcbbb81f6cb9e51c309903a610e951c69ebaae), [register.ts](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-e2e17828402e48441259e515015cac450bb7966aec49f53e56f2a2b76e531349), and [notifications.router.ts](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-43e37c2b116a422fe013e16448a53cd82309419b029ada5db54f42cc20f78f58).
- Add v2 middleware for AppCheck-only and JWT/AppCheck auth and JWT utilities enforcing `MAX_JWT_METADATA_SIZE=1024` in [auth.ts](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-a9d14facf2b63e9cea5e54d533e749002bc7f9783f58a8851e2231a56c658cbc) and [jwt.ts](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-995f6340d0a8bbea41f06a750751e1d0e1b6778460aef5cf32a438d6853510fb); add webhook auth middleware in [xmtpWebhookAuth.ts](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-6c722e17c49bd8bcb91ec97edc85acef53281d41e6444fec913d5a5963a6f957).
- Add config loader with required `XMTP_NOTIFICATION_SECRET`, `JWT_SECRET`, and `NOTIFICATION_SERVER_URL`, and replace direct `process.env` access across server, instrumentation, error handler, healthcheck, and notifications client in [config.ts](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-c3095d5010e65c52737a98a5d618ea24049ebe90c8470752426081d70ed6e012), [index.ts](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80), [instrumentation.ts](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-1d5bbd60ea57ea5431ebe7f03e6a47d12e205b27e1b04d9e2a046fafde4bbb2c), [errorHandler.ts](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-700a0272fd9808834a4a61705cc95a03bb8523f08fa719efc93341d21175a3cb), [healthcheck.ts](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-fea3beb5ce04504c8c5f2721e74093604bfbba9e3daa8ac7892036b9d10593ee), and [notifications client](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-5e63127ae8931e94ca574770a10a816072cfbd2e59a476e247d870161400f59c).
- Update notification services and types to support v1/v2 payloads, make `encryptedMessage` optional, redact APNS auth header in logs, and reject FCM in [apns-push.service.ts](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-40f7eed61750bd835838192a9727a8078b3a2637d14d4103739840f46efa3184), [notifications-types.ts](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-20b29a72905be43d8f589414d266a3bb48c5970db7d21df62c56e1c383140272), and [push-notification.service.ts](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-76ecd70164eb9efd61dd676f88cd527ee85cbcfdad9de40ffd5ca2ccb0330c65).
- Extend v2 invites decode to include `conversationExpiresAt`, `expiresAt`, and `expiresAfterUse`, and update proto and tests in [invite.proto](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-b92c079831e9fecde880360e2625726b6048979b7c3943f8918d7c76d9e6bb6c) and [decode-invite-slug.ts](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-2d0c3ad748e88210424de2d0b48e5626544596ebf6ccc022f69a51d90b5faa43).
- Adjust rate limits to 1000/5m general and 100/5m for auth in [rateLimit.ts](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-b3f285f74315ca0a0594a1319122cf3b0a5b15dcab60014c1f8bd6896818a8db); add constants `MAX_PUSH_FAILURES=10` and `MAX_JWT_METADATA_SIZE=1024` in [constants.ts](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-68d9908e0078a54ad38ca074333ba4136ae93fc5f81c1106b0d116f90a00f5e1).
- Bump Prisma to 6.17.0 and align schema/code; add tsconfig path alias for `@prisma-zod/*` and minor refactors in v1 handlers and tests.

#### 📍Where to Start
Start with the shared XMTP webhook entrypoint `handleXmtpNotification` in [webhook-handler.ts](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-9ee12a996c2234bd60a6517d734ac22eb96ef397eb6f6b763710d4d8047df9c7), then review the v2 device registration flow in [register.ts](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-e2e17828402e48441259e515015cac450bb7966aec49f53e56f2a2b76e531349) and the v2 auth/token flow in [generate-token.ts](https://github.com/ephemeraHQ/convos-backend/pull/141/files#diff-efae9276d245e77d478e0961baa0a511ec789b80927164138b3523f31781fb70).

#### Changes since #141 opened

- Updated test data in V2 Invite API test case [9c17fdd]
- Added runtime validation for JWT_SECRET configuration [11ab9dc]
- Fixed IS_PRODUCTION and IS_DEVELOPMENT constants evaluation [11ab9dc]
----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 11ab9dc. 28 files reviewed, 46 issues evaluated, 46 issues filtered, 0 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>proto/invite/v2/invite.proto — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 28](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/proto/invite/v2/invite.proto#L28): The field `expiresAfterUse` is declared as a plain `bool` in proto3 (`bool expiresAfterUse = 9;`). In proto3, scalar `bool` fields do not have presence, meaning there is no way to distinguish an "unset" value from an explicitly set `false`. If downstream logic needs to differentiate between "not provided" and `false` (e.g., for partial updates, defaults, or backward compatibility), this schema will silently conflate the two states. To preserve three-state semantics (unset/true/false), use `google.protobuf.BoolValue` or wrap the field in a `oneof`. Otherwise, explicitly document and enforce that the default `false` is always correct when the field is omitted. <b>[ Low confidence ]</b>
</details>

<details>
<summary>src/api/shared/notifications/services/apns-push.service.ts — 0 comments posted, 4 evaluated, 4 filtered</summary>

- [line 109](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/shared/notifications/services/apns-push.service.ts#L109): Unhandled error path when generating the APNS JWT: `getJwtToken()` is invoked at line 109 without any try/catch. If `jwt.sign()` throws (e.g., invalid `privateKey`, malformed key format, unsupported algorithm, or runtime crypto error), `sendPushNotification` will throw and return a rejected promise instead of the advertised `{ success: boolean; error?: string }` shape. This violates the method's contract and yields an ungraceful failure for callers. <b>[ Out of scope ]</b>
- [line 121](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/shared/notifications/services/apns-push.service.ts#L121): Potential multiple resolution of the Promise due to multiple event handlers calling `resolve`: `client.on('error')` (lines 121–128), `request.on('error')` (lines 236–248), and `request.on('end')` (lines 181–234) each call `resolve`. Without guarding against re-entrancy (e.g., a `settled` flag) or using `.once` listeners, there is a risk that more than one of these handlers fires, causing double application of side effects (e.g., multiple `client.close()` calls) and violating the at-most-once completion invariant, even though subsequent `resolve` calls are ignored by the Promise. <b>[ Out of scope ]</b>
- [line 145](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/shared/notifications/services/apns-push.service.ts#L145): Sensitive data leakage in logs: the code logs the entire `payload` object at lines 141–149. Given the accepted type `AnyNotificationPayloadWithJWT`, the payload may contain a JWT or other secrets specific to the notification, which will be exposed in logs. While the `authorization` header is redacted via `safeHeaders`, the payload is not redacted. Redact or omit sensitive fields in `payload` before logging. <b>[ Out of scope ]</b>
- [line 152](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/shared/notifications/services/apns-push.service.ts#L152): No timeout or abort is configured for the HTTP/2 client or request. If APNS or the network stalls (no 'response', 'end', or 'error'), the Promise never resolves and the client connection can linger. This causes hanging calls and potential resource leakage. Add a request or client timeout and abort/close handlers to guarantee a terminal state. <b>[ Out of scope ]</b>
</details>

<details>
<summary>src/api/shared/notifications/services/notifications-types.ts — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 6](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/shared/notifications/services/notifications-types.ts#L6): The type change making `ProtocolNotificationData.encryptedMessage` optional (`encryptedMessage?: string`) loosens the contract. Any downstream code that assumes this field is present and attempts to parse or decrypt it without an existence check could encounter runtime failures (e.g., using it as a string). Ensure that all consumers of `ProtocolNotificationData` handle the `undefined` case gracefully by checking for existence before use, and that any emitted artifacts based on this field remain valid when it is absent. <b>[ Low confidence ]</b>
</details>

<details>
<summary>src/api/shared/notifications/webhook-handler.ts — 0 comments posted, 6 evaluated, 6 filtered</summary>

- [line 15](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/shared/notifications/webhook-handler.ts#L15): Module-level initialization of `notificationClient` relies on `createNotificationClient()` which now does not validate `NOTIFICATION_SERVER_URL`. If the env var is missing, subsequent calls (e.g., `deleteInstallation`, `unsubscribe`) will fail at runtime with transport errors. This should be validated pre-effect at module load or handled gracefully during request handling by guarding against undefined base URL and producing a clear error. <b>[ Invalidated by documentation search ]</b>
- [line 49](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/shared/notifications/webhook-handler.ts#L49): Potential runtime crash due to assuming `req.log` exists on the Express `Request`. The code calls `req.log.*` in multiple places (e.g., `req.log.error` on parse failure). The imported `Request` type from `express` does not include `log`. If a logging middleware has not attached `log` to the request object for this route, accessing `req.log` will throw a `TypeError` at runtime. Either refine the request type to include `log`, add a guard before using it, or ensure middleware attaches `log` on all paths to this handler. <b>[ Low confidence ]</b>
- [line 139](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/shared/notifications/webhook-handler.ts#L139): Potential crash due to missing null check on `identity` in the v1 path. In `handleXmtpNotification`, the code accesses `identity.xmtpId` when building the push notification payload. However, `identity` is loaded via `include: { identity: true }` from `prisma.identitiesOnDevice.findUnique` without any explicit guarantee that the related `identity` exists. If `identity` is `null` or missing, accessing `identity.xmtpId` will throw a runtime error. Add an explicit check that `identity` is present (and contains a valid `xmtpId`) before dereferencing it, and handle the error path gracefully. <b>[ Low confidence ]</b>
- [line 186](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/shared/notifications/webhook-handler.ts#L186): `cleanupFailedInstallation` increments `pushFailures` on the `device` table (lines 186–198), while the v2 code uses `deviceRegistration` to track `pushFailures` and `disabled` state. This split accounting leads to inconsistent state: failures on v1 cleanup go to `device.pushFailures`, but v2 notification handling reads and updates `deviceRegistration.pushFailures` and `deviceRegistration.disabled`. As a result, thresholds may be enforced on `deviceRegistration` but not on `device`, and vice versa. Additionally, `cleanupFailedInstallation` does not perform any conditional disable when exceeding `MAX_PUSH_FAILURES`, so a device may accumulate failures without ever being disabled in the `device` model. This divergence can cause mismatches in skip/disable decisions and stale device state. <b>[ Low confidence ]</b>
- [line 225](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/shared/notifications/webhook-handler.ts#L225): The pre-check in `handleV2Notification` skips sending when `client.device.disabled` or `client.device.pushFailures >= MAX_PUSH_FAILURES` (lines 225–233), but it does not enforce the disablement when the threshold is exceeded. In failure-handling paths, the code increments `pushFailures` and atomically sets `disabled = true` when the threshold is reached, but in this early-return path there is no state update — it just returns `{ success: false }`. This creates inconsistent behavior: a device with `pushFailures` already at or above the threshold will be skipped but not marked disabled, leaving inconsistent flags across devices depending on which path was last executed. <b>[ Low confidence ]</b>
- [line 248](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/shared/notifications/webhook-handler.ts#L248): When APNS is not configured, `handleV2Notification` returns `{ success: false }` without recording a push failure or updating any device state. This silently drops notifications and prevents the failure counter and disable logic from ever being applied for the affected device. Concretely, at lines 248–251, if `createApnsService()` returns `null`, the code logs an error and returns early; it does not increment `client.device.pushFailures`, set `lastFailureAt`, or conditionally set `disabled` when reaching `MAX_PUSH_FAILURES`. This leads to devices repeatedly failing to send due to global misconfiguration without ever being disabled or tracked, violating the intended invariant that repeated failures should be accounted for and eventually cause disablement. <b>[ Invalidated by documentation search ]</b>
</details>

<details>
<summary>src/api/v1/notifications/handlers/register-installation.ts — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 64](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/v1/notifications/handlers/register-installation.ts#L64): Ambiguous request classification: the handler first attempts to parse the body against `legacyRegistrationSchema` and routes to `handleLegacyRegistration` if it succeeds. Zod's default object parsing permits unknown keys (non-strict), so a request that includes both legacy fields (e.g., `installationId`, `deliveryMechanism`) and current fields (e.g., `deviceId`, `pushToken`, `installations`) will still satisfy `legacyRegistrationSchema`. In such cases, the function will incorrectly prefer the legacy path and ignore the newer `installations` array, leading to unexpected behavior and an externally visible contract mismatch on response shape. This is a runtime logic bug because inputs that include superset fields are reachable, and there is no guard preventing dual-format bodies. <b>[ Low confidence ]</b>
</details>

<details>
<summary>src/api/v2/device/handlers/register.ts — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 27](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/v2/device/handlers/register.ts#L27): `req.log` is used throughout the `register` handler without a visible guard or guarantee that the property exists on the Express `Request` object. In standard Express, `req.log` is not defined; unless middleware augments `req`, this will throw at runtime on the first use (e.g., `req.log.warn(...)`). This affects all logging calls: `warn`, `info`, and `error` within the handler. To prevent runtime crashes, either ensure a middleware always attaches a logger to `req` before this handler runs, or add a fallback/guard (e.g., use a module-level logger or check `req.log` presence). <b>[ Low confidence ]</b>
- [line 56](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/v2/device/handlers/register.ts#L56): Empty string input for `pushToken` cannot clear an existing token during update despite the comment stating normalization of empty string to `null`. The schema transforms empty string to `undefined` at parse time (`transform((val) => (!val || val.trim() === "" ? undefined : val))`), and the update logic only applies `pushToken` when `body.pushToken !== undefined`. As a result, sending an empty string will cause `pushToken` to be omitted from `updateData`, leaving any existing token unchanged. This contradicts the in-code comment "Normalize: undefined or empty string -> null." To fix: either remove the transform and handle empty string in the update branch (`pushToken === "" -> null`), or keep the transform but set `updateData.pushToken = null` explicitly when the client intends to clear it (e.g., introduce a distinct field or logic to interpret omission vs. explicit clear). <b>[ Invalidated by documentation search ]</b>
- [line 100](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/v2/device/handlers/register.ts#L100): When clearing the old device’s registration upon push token transfer, the code also forcefully sets `pushTokenType` to `"apns"` and `apnsEnv` to `null` (`data: { pushToken: null, pushTokenType: "apns", apnsEnv: null }`). The comment only states "Clear the push token from the old device to avoid unique constraint violation" and does not mention altering type/environment, so this change can cause unintended data loss for the old device (e.g., losing its previous `apnsEnv` setting). If the model requires non-null `pushTokenType`, a safer approach is to leave other fields unchanged (only set `pushToken: null`) or set `pushTokenType` to a neutral value only if necessary, with explicit documentation clarifying the intended invariant. <b>[ Low confidence ]</b>
</details>

<details>
<summary>src/api/v2/invites/handlers/decode-invite-slug.ts — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 95](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/v2/invites/handlers/decode-invite-slug.ts#L95): `decodeInviteSlug` calls `recoverPublicKey(signedInvite)` but discards the result and performs no signature/authenticity verification of `signedInvite.payload`. As long as the `signature` is present with length 65, a forged or invalid signature and payload will be accepted, and the function will return a `DecodedInvite` without validating that the payload was actually signed by a trusted key (e.g., comparing the recovered public key to an expected signer or verifying the signature over the canonicalized payload). This is a runtime security bug: invalid or tampered invites can pass as valid. To fix, perform an actual signature verification (e.g., use `secp256k1.ecdsaVerify` or compare the recovered public key to the expected signer derived from `creatorInboxId` or another trusted source) and fail if verification does not pass. <b>[ Out of scope ]</b>
- [line 145](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/v2/invites/handlers/decode-invite-slug.ts#L145): `timestampDate(decoded.conversationExpiresAt).toISOString()` and `timestampDate(decoded.expiresAt).toISOString()` can throw at runtime if the input is truthy but invalid (e.g., a malformed timestamp) and `timestampDate` returns an invalid `Date`. `Date.prototype.toISOString()` throws a RangeError for invalid dates. There is no validation/try-catch around these conversions, so a malformed but truthy input can crash the handler. <b>[ Low confidence ]</b>
- [line 150](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/v2/invites/handlers/decode-invite-slug.ts#L150): `expiresAfterUse` is not normalized to `null` when absent, unlike other nullable fields (`name`, `description`, `imageURL`, `conversationExpiresAt`, `expiresAt`) which use `?? null`. If `decoded.expiresAfterUse` is `undefined`, the response will contain `expiresAfterUse: undefined`. If the `DecodeInviteSlugResponse` contract expects `null` for missing values (as suggested by the other fields’ handling), this creates an inconsistent payload and may break consumers relying on explicit `null` for "absent" fields or strict JSON schemas that disallow `undefined`. <b>[ Low confidence ]</b>
</details>

<details>
<summary>src/api/v2/notifications/handlers/subscribe.ts — 0 comments posted, 7 evaluated, 7 filtered</summary>

- [line 13](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/v2/notifications/handlers/subscribe.ts#L13): `topic` is validated as `z.string()` with no constraints; empty strings or excessively long/invalid topic names are accepted. These are likely rejected by the notification server, resulting in a 500 instead of a 400. Add schema constraints (e.g., `.min(1)` and domain-specific validation) to prevent invalid topics reaching downstream. <b>[ Low confidence ]</b>
- [line 14](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/v2/notifications/handlers/subscribe.ts#L14): `hmacKeys` is validated as `z.array(...)` with no `.min(1)` constraint. An empty array is accepted, producing `subscriptions` with `hmacKeys: []`. This is likely invalid for the notification server and will cause a remote error, reported as 500. Enforce `.min(1)` to correctly reject invalid client input as 400. <b>[ Low confidence ]</b>
- [line 16](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/v2/notifications/handlers/subscribe.ts#L16): `thirtyDayPeriodsSinceEpoch` is validated as a generic `z.number()` without ensuring it is an integer, non-negative, or within a reasonable range. Fractional or negative values are accepted and will be forwarded to the notification server, likely causing a remote validation error and a 500 response. The schema should enforce integer (`z.int()`), `min(0)`, and potentially reasonable bounds to return a 400 for invalid client input. <b>[ Low confidence ]</b>
- [line 28](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/v2/notifications/handlers/subscribe.ts#L28): Missing configuration guard for `NOTIFICATION_SERVER_URL` can cause runtime failure at module initialization or later remote-call failures. `createNotificationClient()` is invoked at module load (singleton `notificationClient`), but if `NOTIFICATION_SERVER_URL` is undefined or invalid, `createConnectTransport({ baseUrl: NOTIFICATION_SERVER_URL })` may throw or produce a client that fails on first use. The previous explicit check was removed in the diff, so startup may proceed with a broken client and only fail during handling requests. <b>[ Invalidated by documentation search ]</b>
- [line 37](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/v2/notifications/handlers/subscribe.ts#L37): `req.log` is assumed to exist, but `express.Request` does not guarantee a `log` property. If a request arrives without the logging middleware attached, the first use at `req.log.info(...)` will throw, and the catch block also calls `req.log.error(...)`, which will throw again, potentially resulting in an unhandled exception and no response sent. A guard or fallback logger is required before use on both success and error paths. <b>[ Low confidence ]</b>
- [line 81](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/v2/notifications/handlers/subscribe.ts#L81): Hex key validation allows any-length hex strings via `z.string().regex(/^[0-9a-fA-F]+$/)`, but many hex-to-bytes converters (including `hexToUint8Array`) expect an even-length input. An odd-length hex string will cause `hexToUint8Array(key.key)` to throw at runtime and be reported as a 500 error instead of a 400 validation error. Enforce even length in the schema or handle conversion errors to return a client error. <b>[ Invalidated by documentation search ]</b>
- [line 92](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/v2/notifications/handlers/subscribe.ts#L92): Push token type handling assumes only two values: `'apns'` maps to `case: 'apnsDeviceToken'`, and all other values map to `case: 'firebaseDeviceToken'`. If `device.pushTokenType` is absent, unrecognized, or a different APNS variant (e.g., sandbox), this conditional will mislabel the delivery mechanism and may cause remote rejection. The schema should validate `pushTokenType` against allowed values, and the mapping should cover all supported cases explicitly with a clear error when unsupported. <b>[ Low confidence ]</b>
</details>

<details>
<summary>src/api/v2/notifications/handlers/unregister.ts — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 38](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/v2/notifications/handlers/unregister.ts#L38): Both `unregister()` and `unsubscribe()` pass `installationId: clientId` to the notification service (`deleteInstallation` and `unsubscribe`). In v1 paths and example usage, `installationId` denotes an XMTP installation identifier distinct from a device/client ID (e.g., `xmtpInstallationId`). If the notification service expects an installation identifier rather than a client identifier, these calls may target the wrong resource or no-op, leading to failed cleanup/unsubscription. Verify the contract: if v2 requires passing `clientId` instead of XMTP installation id, then the server must map accordingly; otherwise, translate `clientId` to the correct `installationId` before invoking the notification client. <b>[ Low confidence ]</b>
- [line 38](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/api/v2/notifications/handlers/unregister.ts#L38): In `unregister()`, cleanup operations are performed in two separate calls: `notificationClient.deleteInstallation()` followed by `prisma.clientIdentifier.delete()`. If the first call succeeds (external installation removed) but the second fails (database deletion), the system ends up in an inconsistent state: the client record remains locally while the remote installation has been deleted. Consider wrapping both operations in a single transactional pattern with compensating logic (e.g., delete local first then remote, or use a transaction with best-effort rollback where feasible) to maintain invariants. At minimum, ensure partial failure is handled by explicitly reconciling state to avoid dangling records. <b>[ Low confidence ]</b>
</details>

<details>
<summary>src/config.ts — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 31](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/config.ts#L31): Startup crash risk: `JWT_SECRET_BYTES = new TextEncoder().encode(JWT_SECRET)` will throw a `TypeError` at module initialization if `process.env.JWT_SECRET` is undefined, because `TextEncoder.encode` requires a string. If undefined configuration is deployment-invalid, it’s acceptable to terminate at initialization, but as written there’s no explicit guard or clearer error, and the throw will be a generic TypeError. Consider adding an explicit check with a descriptive error and failing fast before external effects, e.g., `if (!JWT_SECRET) throw new Error("JWT_SECRET is required")`. <b>[ Low confidence ]</b>
- [line 32](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/config.ts#L32): `NOTIFICATION_SERVER_URL` is exported directly from `process.env` and can be `undefined`. Without validation at initialization, downstream code may assume it is a string and attempt to use it in URL construction or network calls, causing runtime errors (e.g., `TypeError: Failed to construct 'URL'`). If this configuration is required for correct operation, it should be validated as part of process/service initialization to fail fast before side effects. <b>[ Invalidated by documentation search ]</b>
</details>

<details>
<summary>src/instrumentation.ts — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 31](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/instrumentation.ts#L31): The `resourceFromAttributes` call may include an invalid attribute value: `"deployment.environment": process.env.ENV`. When `process.env.ENV` is undefined (which is reachable with no guards), the resulting attribute value will be `undefined`. OpenTelemetry `Attributes` do not permit `undefined` values, and emitting an attribute with an undefined value can lead to malformed resource data and potential exporter/runtime errors. This violates the requirement that emitted artifacts be valid under all reachable inputs. Guard this value (e.g., omit the key when undefined or provide a default such as `process.env.NODE_ENV`) so the attributes passed to `resourceFromAttributes` are always valid. <b>[ Out of scope ]</b>
</details>

<details>
<summary>src/middleware/errorHandler.ts — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 16](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/middleware/errorHandler.ts#L16): Dereferencing `req.log.error` without a guard can throw at runtime if `req.log` is not present on the `Request` object or if `req.log.error` is not a function. The standard Express `Request` type does not include a `log` property; it is commonly added by logging middleware (e.g., `pino-http`). If that middleware is absent or fails, `req.log` will be `undefined`, causing `TypeError: Cannot read properties of undefined (reading 'error')` in the error handler, which prevents sending any response for the original error. Add a defensive check and fallback (e.g., to `console.error`) to ensure the handler never throws during logging. <b>[ Out of scope ]</b>
- [line 27](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/middleware/errorHandler.ts#L27): Serializing `err.details` directly in the `AppError` branch can throw at runtime if `details` contains non-JSON-serializable values (e.g., `BigInt`, `Symbol`, or circular references). `res.json(...)` will attempt to stringify the response; if `details` is problematic, the error handler itself will throw, resulting in no response sent and potentially an unhandled exception cascade. Since `details` is typed as `unknown` and can be any value, guard or sanitize it before serialization (e.g., map to a safe shape, stringify with a safe replacer, or omit when non-serializable). <b>[ Out of scope ]</b>
</details>

<details>
<summary>src/middleware/rateLimit.ts — 0 comments posted, 4 evaluated, 4 filtered</summary>

- [line 6](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/middleware/rateLimit.ts#L6): The `limit` option in `rateLimitMiddleware` may be ignored or misinterpreted depending on the installed version of `express-rate-limit` (some versions expect `max` instead of `limit`). If the package version expects `max`, using `limit` could cause unintended defaults or behavior (e.g., different rate limits than configured). Ensure the option name matches the installed library version, or set both `limit`/`max` according to documented compatibility. <b>[ Low confidence ]</b>
- [line 8](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/middleware/rateLimit.ts#L8): The `standardHeaders` option is set to `"draft-8"` in `rateLimitMiddleware`. Depending on the installed version of `express-rate-limit`, this string value may be unsupported, which can cause either a runtime error during middleware creation or silently fall back to unexpected behavior (missing or incorrectly formatted `RateLimit` headers). Validate that the package version supports this exact literal value, or use the documented boolean/union values supported by the installed version. <b>[ Out of scope ]</b>
- [line 14](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/middleware/rateLimit.ts#L14): The `limit` option in `authRateLimitMiddleware` may be the wrong option name for certain versions of `express-rate-limit` (which use `max`). If the installed version expects `max`, using `limit` can result in unintended default limits or ignored configuration, potentially over-throttling or under-throttling auth requests. Confirm and match the option name to the library version. <b>[ Low confidence ]</b>
- [line 16](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/middleware/rateLimit.ts#L16): The `standardHeaders` option is set to `"draft-8"` in `authRateLimitMiddleware`. If the installed `express-rate-limit` version does not support this exact value, middleware initialization may throw or headers may be missing/malformed at runtime, reducing client visibility into rate limits. Verify version support for `"draft-8"` or use the documented values for the installed version. <b>[ Low confidence ]</b>
</details>

<details>
<summary>src/middleware/v2/auth.ts — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 36](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/middleware/v2/auth.ts#L36): The middleware conflates server-side initialization/configuration errors with client authentication errors. If `verifyAppCheckToken` throws due to failures in `getFirebaseApp()` or `getAppCheck(app)` (e.g., Firebase not initialized, misconfiguration, transient backend error), the `catch` block treats it as an invalid client token and responds with `401` (`res.status(401).json({ error: "Invalid AppCheck token" })`). This misclassification masks operational issues as client faults, complicating troubleshooting, and may cause inappropriate client behavior (e.g., retrying with different tokens). To fix, differentiate between known token verification failures (e.g., explicit "invalid token" errors) and infrastructure/configuration errors, returning `401` only for invalid/missing tokens and `5xx` for server-side failures, or propagate via `next(error)` for centralized handling. <b>[ Invalidated by documentation search ]</b>
- [line 59](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/middleware/v2/auth.ts#L59): If both `X-Firebase-AppCheck` and `X-Convos-AuthToken` headers are present and the AppCheck token is invalid, the middleware rejects the request with 401 without attempting JWT verification. This causes valid requests (with a valid JWT) to be denied if an AppCheck header is present but invalid. The intent comment says "Try AppCheck first" but does not state that AppCheck should be exclusive. A more robust approach is to fall back to JWT verification when AppCheck verification fails but a JWT is also provided. <b>[ Low confidence ]</b>
- [line 66](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/middleware/v2/auth.ts#L66): All AppCheck verification errors are mapped to a 401 Unauthorized response (`res.status(401).json({ error: "Invalid AppCheck token" })`). This conflates genuine authentication failures with service/internal errors (e.g., network or Firebase service outages). Internal/operational errors should return a 5xx to accurately communicate server failure and avoid misleading clients. <b>[ Low confidence ]</b>
</details>

<details>
<summary>src/middleware/xmtpWebhookAuth.ts — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 23](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/middleware/xmtpWebhookAuth.ts#L23): The middleware dereferences `req.log.error` without any guard that `req.log` exists on the Express `Request`. In many deployments, `req.log` is not a standard property and will be `undefined` unless a logging middleware augments the request. If `req.log` is missing, calls like `req.log.error(...)` will throw a `TypeError` at runtime, preventing the intended error response from being sent and potentially causing unhandled exceptions. This occurs on multiple lines where logging is used: e.g., at the configuration error catch, missing header case, and invalid header case. <b>[ Low confidence ]</b>
</details>

<details>
<summary>src/notifications/client.ts — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 16](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/notifications/client.ts#L16): createNotificationClient no longer validates `NOTIFICATION_SERVER_URL`. With the previous guard removed, `createConnectTransport` is called with `baseUrl` set to `undefined` when the environment variable is missing. This misconfiguration will only surface later at request time with opaque transport errors. Given that `createNotificationClient()` is invoked at module top-level in multiple places (e.g., `src/api/shared/notifications/webhook-handler.ts` and `src/api/v2/notifications/handlers/*`), this violates the invariant that required configuration should be validated pre-effect during initialization. Add an explicit check that `NOTIFICATION_SERVER_URL` is defined and fail fast before creating the client. <b>[ Invalidated by documentation search ]</b>
- [line 226](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/notifications/client.ts#L226): Potential crash in v2 path if `client.device` is missing. In `handleV2Notification`, the code accesses `client.device.disabled`, and later constructs `deviceForApns` using multiple properties from `client.device`. The record is fetched with `include: { device: true }`, but there is no explicit guarantee that the related device exists. If `client.device` is `null`, these property accesses will throw at runtime. Add an explicit presence check (and an error path or cleanup) before dereferencing `client.device`. <b>[ Low confidence ]</b>
</details>

<details>
<summary>src/utils/v2/jwt.ts — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 19](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/utils/v2/jwt.ts#L19): The JWT payload validation schema (`v2JWTPayloadSchema`) accepts any non-empty string (including an empty string, since `z.string()` allows empty by default) for `deviceId`, while the token generation entrypoint enforces a UUID via `generateTokenRequestSchema`. This contract mismatch allows verification to accept tokens with a `deviceId` that would never be produced by the generator (e.g., empty string or non-UUID string), potentially leading to downstream assumptions being violated (handlers might assume `deviceId` is a UUID). Align the verification schema with generation constraints (e.g., `z.string().uuid()`). <b>[ Low confidence ]</b>
- [line 76](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/utils/v2/jwt.ts#L76): All JWT verification errors are mapped to a 401 Unauthorized (`throw new AppError(401, "Invalid or expired token", verifyError)`), which conflates malformed/invalid tokens with internal errors (e.g., misconfigured `JWT_SECRET_BYTES` or jose runtime errors). This can mislead clients and hinder diagnostics; internal errors should be surfaced as 5xx while invalid tokens should remain 401. <b>[ Invalidated by documentation search ]</b>
- [line 90](https://github.com/ephemeraHQ/convos-backend/blob/11ab9dc39dbcedc544e2f089e442f3d43b32d0c5/src/utils/v2/jwt.ts#L90): The helper `isNotificationExtensionOnlyToken` accesses `payload.metadata` without guarding that `payload` itself is non-null/defined. The expression `payload.metadata?.notificationExtensionOnly === true` will throw at runtime if `payload` is `null` or `undefined`. Since TypeScript types are erased at runtime and there is no explicit guard, passing an invalid payload (e.g., from deserialization or an external caller) will cause a crash. A safe alternative is to use `payload?.metadata?.notificationExtensionOnly === true` or validate `payload` before property access. <b>[ Invalidated by documentation search ]</b>
</details>

</details>

## Summary by CodeRabbit

* **New Features**
  * v2 API: token issuance, device register/unregister/subscribe/unsubscribe endpoints, v2 auth and AppCheck, shared webhook mounted for v1/v2.
* **Refactor**
  * Centralized XMTP notification processing and unified webhook routing; shared notification types.
* **Chores**
  * DB migrations for device/client models and unique push-token constraint; new config and V2 JWT utilities; dependency reorganization.
* **Tests**
  * Tests updated to use shared notification modules and include new invite expiry fields.
* **Bug Fixes**
  * Adjusted API rate limits and auth-specific limiter.


src/api/v2/auth/handlers/generate-token.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    // Check if device is disabled
+    if (client.device.disabled) {
+      res.status(403).json({ error: "Device is disabled" });
+      return;
+    }
+
+    // Generate JWT
+    const token = await createV2JwtToken({
+      clientIdentifier: body.clientIdentifier,
```
neekolas: What do we use the clientIdentifier for? It seems like the `deviceId` is the only thing that is trustworthy. 

The clientIdentifier is generated by the client and can't be verified, right? If we tried to use it for sybil resistance in the gateway we'd get hosed.

What if we left it out entirely and treated the deviceId as the only identifier?

src/utils/v2/jwt.ts
```diff
@@ -0,0 +1,88 @@
...
+
+export const verifyV2JwtToken = async (args: { token: string }) => {
+  if (!process.env.JWT_SECRET) {
+    throw new AppError(500, "JWT_SECRET is not configured");
+  }
+
+  const { data: verified, error: verifyError } = await tryCatch(
+    jose.jwtVerify(
+      args.token,
+      new TextEncoder().encode(process.env.JWT_SECRET),
```
neekolas: For [performance](https://basharat720.medium.com/the-hidden-performance-trap-in-node-js-process-env-isnt-what-you-think-fb8f1cca2487) you might want to move this into a config or variable on startup rather than reading from the env directly on each usage.

src/api/shared/notifications/webhook-handler.ts
```diff
@@ -0,0 +1,311 @@
...
+    // Log the notification for debugging
+    req.log.info(
+      {
+        contentTopic: notification.message.content_topic,
+        installationId: notification.installation.id,
+      },
+      "received notification",
+    );
+
+    // Verify the authorization header
```
neekolas: Seems like this could be a middleware

src/api/v2/auth/handlers/generate-token.ts
```diff
@@ -0,0 +1,74 @@
...
+      deviceId: body.deviceId,
+      expirationTime: "15m", // Short-lived for app-generated requests
+      metadata: {
+        gatewayAuthorized: true,
```
neekolas: When would we be creating tokens where this would be set to false? If there isn't a case we might not need the field.

src/api/v2/auth/handlers/generate-token.ts
lourou: @neekolas clientIdentifier is mapped to the inboxId locally by Convos iOS. We use clientIdentifier it to route notifications to the correct inbox through this alias which is not bound to XMTP, so that when the push arrives, the NSE maps clientIdentifier to the inboxId, and knows with which key to decode the message.

lourou: You're absolutely right that clientIdentifier can't be trusted for Gateway authentication. We don't use it for that, just for notif routing.

src/utils/v2/jwt.ts
lourou: Thanks for this! I'm now caching the most used environment variables at module load time instead of reading from `process.env`

src/api/v2/auth/handlers/generate-token.ts
lourou: ~Should we keep both here: the JWT should explicitly authorize the deviceId + clientIdentifier pairing?~

After diving further, I'm removing the client identifier from the JWT token!

src/api/shared/notifications/webhook-handler.ts
lourou: Yes 💯 this is done

src/api/v2/auth/handlers/generate-token.ts
lourou: Yes, I've removed it as we don't need to have it here as metadata

PR #194, opened 2026-05-08
Auth API: SIWE upgrade path on /auth/token + Account/AuthMethod

fbac:
## Summary

- Adds optional SIWE (EIP-4361) upgrade on `POST /auth/token`. Successful verify upserts `Account` + `AuthMethod(SIWE)` and embeds `accountId` in the JWT.
- Pure addition: legacy device-only clients (no `siwe` field) keep working unchanged. `V2JWTPayload` widens with optional `accountId`.
- New `POST /auth/nonce` issues HMAC-signed `__Host-`-prefixed cookie + DB-backed atomic single-use `AuthNonce`. Sweep runs every 10 min via `setInterval`.
- New `requireAccount` middleware → **403** (`{ error: "Account required" }`) when JWT lacks `accountId`. 401 would push iOS into a re-auth loop.
- Schema: `Account`, `AuthMethod` (`@@unique([type, externalKey])`), `AuthNonce`. UUIDs server-generated via `gen_random_uuid()`. `pgcrypto` extension declared in migration.
- siwe library wrapper enforces EIP-4361 fields the lib doesn't check: domain, version, nonce, chainId allowlist, uri, expirationTime ≤ now+10min, issuedAt ±5min skew, notBefore.
- `cookie-parser` mounted before v2 router (prerequisite for nonce cookie).
- Pinned `siwe@3.0.0` exact (security-critical dep).
 
## What this unblocks

- `payments-repoint` branch — `accountId` available in JWT for pool to persist `instances.owner_account_id` at claim.
- Composio identifier alignment — `accountId` is the stable `userId` target, replacing `deviceId`.
- Future auth methods (`AuthMethodType` enum + new `verify*` helpers; no schema rewrite).
 
## Out of scope (explicit non-goals)                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                        
  - `XmtpInbox`, `Profile`, agent template/skill/instance schemas.                                                                                                                                                                                                                                      
  - `/credits/*` HTTP routes (next branch).                                                                                                            
  - `/accounts/me/*` routes; `POST /accounts/me/auth_methods` link-while-authed.                                                                                                                                                                                                                        
  - Per-instance agent API keys (separate hardening branch).                                                                                                                                                                                                                                            
  - Account merging / cold-login dedup (UX problem; schema does not preclude).                                                                                                                                                                                                                          
  - Account `disabled` / soft-delete; `AuthMethod.status` (revocation theater under current unique constraint).
 
## Security notes

[...]

PR #195, opened 2026-05-08
feat: add agent template foundation utilities [STACKED ON #194]

saulmc:
## Agent Template Foundation Utilities

Foundation layer for the agent-templates feature stack. Builds on Borja's auth PR #194 — `Account` uses UUID primary keys; this branch adds the supporting utilities, admin seed, and auth-middleware upgrades agent-templates needs.

### What this PR adds

| **Slug hashing** | `src/utils/slug-hash.ts` | Byte-verbatim port of pool's `buildSlug` / `slugHash` / `isHashedSlug` — SHA1 → base36 → 5-char hash |
| **Reserved slugs** | `src/utils/reserved-slugs.ts` | 7 reserved words, `validateSlug()`, `isReservedSlug()` |
| **Auth helpers** | `src/utils/auth-helpers.ts`, `src/utils/constants.ts` | `getEffectiveOwnerId(res)` unified identity; `ADMIN_ACCOUNT_ID` constant |
| **Auth middleware** | `src/middleware/agentAuth.ts` | `authOrAgentApiKeyAuth` (JWT or `X-Agent-API-Key`); `requireAccount` for ownership gating |
| **Admin seed** | `prisma/migrations/20260508140000_seed_admin_account/` | Inserts the deterministic-UUID admin account row |
| **Router shell** | `src/api/v2/agent-templates/agent-templates.router.ts` | Empty router gated on `XMTP_ENV !== "production"` |
| **PostHog dep** | `package.json`, `.env.example` | `posthog-node@5.33.3` + builder env vars |
| **Engines** | `package.json` | Bun ≥ 1.0 |

### Key design choices

- **Account model**: UUID primary keys inherited from `fbac/authentication-api`. The admin row is created by a follow-up migration on this branch.
- **Production guard**: `process.env.XMTP_ENV !== "production"` — the router shell (and everything stacked on it) only mounts in dev/staging.
- **`ADMIN_ACCOUNT_ID`**: Deterministic UUID for the admin seed row; used as the effective owner for API-key-authenticated callers.
- **`getEffectiveOwnerId(res)`**: Handlers call this instead of hardcoding `ADMIN_ACCOUNT_ID`. It returns `res.locals.accountId`, which the auth middleware sets for both JWT and API-key paths.
- **`isApiKeyListener`**: Flag set to `true` when API-key auth is used, which downstream PRs use to bypass per-template ownership guards (admin-like access).
- **`agent-skills` deferred**: The original plan included an `agentSkillsRouter` shell here, but skills work has been deferred (lands in a later milestone). Not included in this stack.

No HTTP endpoints are added — only the building blocks the next three PRs in the stack depend on.

### Stacking

```
fbac/authentication-api (PR #194) ← this PR (#195)
  └─ CRUD (PR #199)
      └─ builder (PR #200)
          └─ create-job + twitter (PR #201)
```

> [!NOTE]
> ### Add agent template foundation utilities including slug helpers, admin account seed, and auth identity flags
> - Adds slug validation and hashed-slug generation utilities in [reserved-slugs.ts](https://github.com/ephemeraHQ/convos-backend/pull/195/files#diff-3efd2fc6492e0e823b477bc3f85d6e9c9a8d2cd785fdd62cdad1c87f9ed0dc1e) and [slug-hash.ts](https://github.com/ephemeraHQ/convos-backend/pull/195/files#diff-a2e1fb80c87864b575c47c8b0d02b39ee01a6c649809f3e35d64370217d10caf), including reserved word detection, format/length checks, and retry-based unique slug building.
> - Seeds a deterministic admin account row (`ADMIN_ACCOUNT_ID`) via a new Prisma migration and exports the UUID as a shared constant.
> - Extends `authOrAgentApiKeyAuth` middleware to set `res.locals.accountId` and `res.locals.isApiKeyListener` on requests authenticated via `X-Agent-API-Key`, and adds a `getEffectiveOwnerId` helper to read the resolved account.
> - Mounts an empty `agentTemplatesRouter` at `/api/v2/agent-templates` only in non-production environments.
> - Adds `posthog-node` as a production dependency and documents new optional env vars (`BUILDER_OPENROUTER_API_KEY`, `EXA_SERVICE_KEY`, `POSTHOG_API_KEY`, etc.) in `.env.example`.
> - Behavioral Change: Node.js engine requirement raised from `>=18` to `>=20`; existing test helpers updated to preserve the seeded admin account row during teardown.
>
> 
> #### Changes since #195 opened
>
> - Introduced config-backed accessor function with test override capability for agent assets API key [eb35046]
> - Refactored agent authentication middleware to use config accessor instead of direct environment variable access [eb35046]
> - Updated authentication middleware tests to use override function instead of environment variable mutation [eb35046]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0f4aa6d.</sup>
> 
>

### DIFF
--- a/src/utils/slug-hash.ts
+++ b/src/utils/slug-hash.ts
@@ -1,0 +1,28 @@
+import crypto from "node:crypto";
+
+// 32-bit fingerprint encoded in base36 fits in 5 chars (5 * log2(36) ≈ 25.85 bits).
+const HASH_LEN = 5;
+const HASHED_SLUG_RE = /\.[0-9a-z]{5}$/;
+
+/**
+ * Stable 5-char base36 hash derived from the agent ID. Used to suffix the
+ * slug (`brewski.x4f9k`) so two skills with the same name don't collide on
+ * URL, and so published pages aren't trivially guessable from the agent name.
+ */
+export function slugHash(id: string): string {
+  const sha = crypto.createHash("sha1").update(id).digest("hex");
+  const n = parseInt(sha.slice(0, 8), 16);
+  return n.toString(36).padStart(HASH_LEN, "0").slice(-HASH_LEN);
+}
+
+/** Combine a base slug with the hash suffix derived from `id`. */
+export function buildSlug(baseSlug: string, id: string): string {
+  return `${baseSlug}.${slugHash(id)}`;
+}
+
+/** True when the slug already has the `.hash` suffix. */
+export function isHashedSlug(slug: string): boolean {
+  return HASHED_SLUG_RE.test(slug);
+}
+
+export { HASH_LEN };
--- a/src/utils/slug-hash.ts
+++ b/src/utils/slug-hash.ts
@@ -3,6 +3,10 @@ import crypto from "node:crypto";
 // 32-bit fingerprint encoded in base36 fits in 5 chars (5 * log2(36) ≈ 25.85 bits).
 const HASH_LEN = 5;
 const HASHED_SLUG_RE = /\.[0-9a-z]{5}$/;
+// 5-char base36 ≈ 67M values. Birthday math: ~1% collision among 1.1k records
+// sharing a base slug, ~50% at 9k. We retry with a fresh ID on collision; cap
+// attempts so a poisoned base slug can't spin forever.
+const MAX_SLUG_ATTEMPTS = 8;
 
 /**
  * Stable 5-char base36 hash derived from the agent ID. Used to suffix the
@@ -25,4 +29,31 @@ export function isHashedSlug(slug: string): boolean {
   return HASHED_SLUG_RE.test(slug);
 }
 
-export { HASH_LEN };
+/**
+ * Reserve a unique hashed slug for a record. Generates a fresh ID per attempt
+ * so a hash collision picks a new suffix instead of clobbering an existing
+ * row. Returns the chosen `{ id, slug }` so the caller persists both.
+ *
+ * Callers should still rely on a DB unique constraint on `slug` as the final
+ * arbiter — `isTaken` and the eventual insert race, and the constraint is
+ * what keeps that race correct.
+ */
+export async function buildUniqueSlug(args: {
+  baseSlug: string;
+  idFactory: () => string;
+  isTaken: (slug: string) => Promise<boolean>;
+}) {
+  const { baseSlug, idFactory, isTaken } = args;
+  for (let i = 0; i < MAX_SLUG_ATTEMPTS; i++) {
+    const id = idFactory();
+    const slug = buildSlug(baseSlug, id);
+    if (!(await isTaken(slug))) {
+      return { id, slug };
+    }
+  }
+  throw new Error(
+    `slug collision: exhausted ${MAX_SLUG_ATTEMPTS} attempts for base slug "${baseSlug}"`,
+  );
+}
+
+export { HASH_LEN, MAX_SLUG_ATTEMPTS };
--- a/.env.example
+++ b/.env.example
@@ -42,6 +42,13 @@ AGENT_ASSETS_API_KEY=
 COMPOSIO_API_KEY=
 COMPOSIO_CONNECTION_CALLBACK_URL=convos://connections/callback
 
+# Builder/template generation (optional)
+BUILDER_OPENROUTER_API_KEY=
+BUILDER_MODEL=
+EXA_SERVICE_KEY=
+POSTHOG_API_KEY=
+POSTHOG_HOST=
+
 # S3 Lifecycle Test Configuration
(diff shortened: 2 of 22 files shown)


saulmc:
* **#197** <a href="https://app.graphite.com/github/pr/xmtplabs/convos-backend/197?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#196** <a href="https://app.graphite.com/github/pr/xmtplabs/convos-backend/196?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#195** <a href="https://app.graphite.com/github/pr/xmtplabs/convos-backend/195?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/xmtplabs/convos-backend/195?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `otr-dev`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

saulmc:
Thanks for the thorough re-review, and for the self-correction on §1.

Confirming the SIWE vars are present at lines 58–69 of `.env.example` on this PR's current head (`8a387c6`):

```
60: SIWE_DOMAIN=convos.app
62: SIWE_URI=https://convos.app
65: SIWE_ALLOWED_CHAIN_IDS=1
69: NONCE_HMAC_SECRET=
```

**Why Macroscope's comment is still visible despite being incorrect:**

PR #195 was originally created via `git reset --soft origin/fbac/authentication-api && git commit`, but the working tree at that moment came from *before* Borja's `812c298` commit landed on `fbac/authentication-api` (which added the SIWE vars + `/account-auth-check` route + `tests/account-auth-check.test.ts`). The resulting squash commit's diff legitimately *did* remove all three of those — Macroscope flagged the env-var portion of that.

The branch has since been re-squashed onto Borja's current tip so the broken commit's diff no longer exists in the PR's history. The cumulative PR diff against base now shows **0 deletions** in `.env.example` (only the builder-vars block as an addition) — verifiable via `gh api repos/xmtplabs/convos-backend/pulls/195/files`:

```
status: modified
additions: 7
deletions: 0
```

Macroscope's comment is anchored to a commit SHA that's been force-pushed away, but GitHub doesn't auto-mark it Outdated.

Resolving the Macroscope thread manually.

src/utils/slug-hash.ts
```diff
@@ -0,0 +1,28 @@
...
+ */
+export function slugHash(id: string): string {
+  const sha = crypto.createHash("sha1").update(id).digest("hex");
+  const n = parseInt(sha.slice(0, 8), 16);
+  return n.toString(36).padStart(HASH_LEN, "0").slice(-HASH_LEN);
+}
+
+/** Combine a base slug with the hash suffix derived from `id`. */
+export function buildSlug(baseSlug: string, id: string): string {
+  return `${baseSlug}.${slugHash(id)}`;
```
neekolas, line 24: The odds of a hash collision on something like this are low. But they're not 0. Had Claude do some back of the envelope math: 1% chance of collision with 1100 assistants sharing the same name. 50% with 9k.

The things we want to avoid is that if you are the unlucky user who hits a slug collision:

1) We don't mangle your assistant state with someone else's
2) We don't put you into some unrecoverable error

Basically, we need to retry on collision and get a new ID s

saulmc: Good catch — added `buildUniqueSlug` in f62cd6a. Generates a fresh `id` per attempt, caps at 8 tries, returns `{ id, slug }` so the caller persists the chosen id. The 5-char hash gives ~67M values, so 8 retries makes the practical chance of exhausting near zero even at the 50%-per-record scale.

Two notes on shape:

1. The util alone is dead code — it has to be wired into the row-creation path. PR #199's `createWithAutoSlug` / `createWithExplicitSlug` need to switch from "let Prisma generate id, then derive URL" to "pick an id with `buildUniqueSlug`, then `prisma.agentTemplate.create({ data: { id, slug, ... } })`". Same in PR #200's `generate-template.ts`. Doing that on #199 next.

2. The DB unique constraint is per-owner (`@@unique([ownerAccountId, slug])`), so the `isTaken` predicate in #199 will need a cross-owner lookup (`findMany({ where: { slug: baseSlug }})` then filter by `slugHash(row.id)`) — otherwise two owners can still mint colliding URLs. Worth considering a Postgres functional unique index on `(slug, slugHashImmutable(id))` cross-owner as a final backstop; will raise that in the #199 thread.

neekolas (approved):

PR #196, opened 2026-05-08
feat: add agent template schema [STACKED ON #195]

saulmc:
## Agent Template CRUD + Schema

**Re-stacked on PR #195** (foundation on `fbac/authentication-api`). AgentTemplate uses UUID `ownerAccountId` referencing the auth PR's Account model.

### What this PR adds

| Route | Method | Auth | Behavior |
|-------|--------|------|----------|
| `/api/v2/agent_templates` | GET | public | Paginated list, keyset cursor, status/category/featured filters |
| `/api/v2/agent_templates/:idOrSlug` | GET | public | Detail by ID or hashed slug, optional `?expand=owner` |
| `/api/v2/agent_templates` | POST | jwt or API key | Create with auto-slug, reserved-slug guard |
| `/api/v2/agent_templates/:id` | PATCH | jwt or API key | Partial update, immutable-field guard |
| `/api/v2/agent_templates/:id` | DELETE | jwt or API key | Hard delete, slug freed for reuse |
| `/api/v2/agent_templates/:id/publish` | POST | jwt or API key | Draft→published, auto-version, `firstPublishedAt` |

### Schema

```prisma
model AgentTemplate {
  id               String        @id
  slug             String
  ownerAccountId   String        @db.Uuid    // UUID FK → Account
  forkedFromId     String?
  agentName        String
  description      String?
  prompt           String        @db.Text
  category         String?
  emoji            String?
  avatarUrl        String?
  tools            String[]      @default([])
  connections      String[]      @default([])
  version          Int           @default(1)
  firstPublishedAt DateTime?
  status           PublishStatus @default(draft)
  featured         Boolean       @default(false)
  createdAt        DateTime      @default(now())
  updatedAt        DateTime      @updatedAt
  // relations, indexes...
}
```

### Key decisions

- **UUID ownerAccountId**: `@db.Uuid` column type, referencing Account.id from auth PR. All handlers use `ADMIN_ACCOUNT_ID` constant.
- **AgentTemplates relation**: `Account.agentTemplates AgentTemplate[]` added alongside `authMethods`.
- **Reserved slugs**: Auto-derived reserved slugs (e.g. `generate` from prompt content) are rejected with 400, not auto-suffixed.
- **Write auth**: `authOrAgentApiKeyAuth` middleware — accepts JWT or `X-Agent-API-Key` header.
- **Keyset cursor**: Encoded `status:createdAt:id` tuple for stable pagination.

### Stacking

```
fbac/authentication-api (PR #194)
[...]

Issue #16, opened 2025-02-14
Profile Picture Upload


lourou:
Currently, Convos lacks profile picture upload functionality. Since Railway doesn't provide native object storage, we need to implement this using AWS or the existing Scaleway's S3-compatible storage solution that's used by the existing backend.

### Requirements
- Endpoint to handle profile picture uploads
- Image processing and compression before storage
   - Only allow images to be uploaded
   - Implement size limits (suggest 8MB max as upload size)
   - Downsize and compress uploaded images (to 250-300KB max)
   - Standardize image dimensions
   - Convert to consistent format (JPEG)

PR #199, opened 2026-05-09
feat: add agent template CRUD with per-account ownership [STACKED ON #195]

saulmc:
## Agent Template CRUD with Per-Account Ownership

Adds the `AgentTemplate` model and full CRUD surface under `/api/v2/agent-templates`. Replaces the merged-too-early PR #196 with proper auth integration baked in from the start.

### What this PR adds

| **Schema** | `prisma/schema.prisma`, migration | `AgentTemplate` model with UUID `ownerAccountId`, `PublishStatus` enum (`draft` / `published` / `unlisted` / `archived`), unique `(ownerAccountId, slug)`, self-referential `forkedFromId` |
| **CRUD handlers** | `handlers/create.ts`, `patch.ts`, `delete.ts`, `publish.ts`, `list.ts`, `detail.ts` | Full lifecycle with ownership + draft-visibility rules |
| **Slug logic** | `lib/resolve-id-or-hashed-slug.ts`, `lib/serialize-agent-template.ts` | Hashed-slug resolution (5-char base36 suffix), serialization |
| **Router** | `agent-templates.router.ts` | Every route chains `authOrAgentApiKeyAuth + requireAccount` |

### Endpoints

| Method | Path | Behavior |
|---|---|---|
| `POST` | `/` | Create draft (auto-slug from `agentName`, or explicit) |
| `GET` | `/` | List with cursor pagination + `?status=`, `?category=`, `?featured=true`, `?owner=` filters |
| `GET` | `/:idOrHashedSlug` | Resolve by UUID, slug, or hashed-slug (`slug.hash`) |
| `PATCH` | `/:id` | Partial update with ownership guard; slug becomes immutable once `firstPublishedAt` is set. `published` / `unlisted` / `archived` → `draft` is allowed (takes the template out of public view; `firstPublishedAt` preserved so the slug stays locked). `draft` → non-draft is rejected with `INVALID_STATUS_TRANSITION` — use `POST /:id/publish` instead. |
| `DELETE` | `/:id` | Hard-delete in any publish state. Forks lose `forkedFromId` (FK `ON DELETE SET NULL`); generations on later branches lose `templateId` the same way. Owner accepts that any cached/bookmarked URL pointing at this template will start 404'ing. |
| `POST` | `/:id/publish` | First-publish stamps `firstPublishedAt` + sets status from `?status=` (default `published`). Re-publish on already-public rows bumps `version` and preserves status (`?status=` ignored). Re-publish on a `firstPublishedAt`-set row currently in `draft` brings it back out of draft, honouring `?status=` or defaulting to `published`. |

### Auth model

- **Every endpoint requires auth** (`authOrAgentApiKeyAuth + requireAccount`). No public discovery surface — unauthenticated callers get 401, regardless of template status.
- **JWT callers** see `published` / `unlisted` / `archived` templates from any owner, plus their own drafts. Status filter (`?status=draft`, etc.) scopes to caller's own templates.
- **API-key listeners** (`X-Agent-API-Key`) see everything (admin-like access) and can mutate any template.
- **Ownership guards** on `PATCH` / `DELETE` / `POST /:id/publish`: 403 unless caller is the owner or an API-key listener. 403s emit `req.log.warn` with `{callerAccountId, templateId, ownerAccountId, action}` for security monitoring.

### Concurrency safety

All mutations use `updateMany` with `WHERE` pinned to the row state observed at validation time. Concurrent publish / edit / delete on the same template surfaces as 409 `TEMPLATE_MODIFIED` (or 404 if the row was deleted out from under us) rather than corrupting state.

### Stacking

PRs #194 (auth API) and #195 (foundation utilities) are now merged to `otr-dev`. Active stack on top of `otr-dev`:

```
otr-dev
  └─ this PR (#199 CRUD)
      └─ PR #204 (generations pipeline)
          └─ PR #205 (twitter integration)
```

> [!NOTE]
> ### Add agent template CRUD endpoints with per-account ownership and slug hashing
> - Adds a new `AgentTemplate` Prisma model with a `PublishStatus` enum and a database migration, establishing relations to `Account` and a self-referential fork relation.
> - Registers six authenticated REST endpoints on the `agentTemplatesRouter`: `GET /`, `POST /`, `PATCH /:id`, `DELETE /:id`, `POST /:id/publish`, and `GET /:idOrHashedSlug`.
> - Implements slug uniqueness per owner with auto-suffix fallback (e.g. `brewski`, `brewski-2`) and cross-owner collision avoidance via a 5-character ID-derived hash appended to slugs.
> - Visibility rules restrict draft/unlisted/archived templates to owners and API key listeners; published templates are publicly listed.
> - Adds a comprehensive integration test suite covering CRUD, auth, ownership guards, pagination, slug behavior, lifecycle transitions, and environment-gated route mounting.
>
> 
> #### Changes since #199 opened
>
> - Added documentation comment to `serializeAgentTemplate` function specifying it as a pure formatter [6bf7286]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 452fa1c.</sup>
> 
>
>

### DIFF
--- a/src/api/v2/agent-templates/handlers/delete.ts
+++ b/src/api/v2/agent-templates/handlers/delete.ts
@@ -1,0 +1,86 @@
+import type { Request, Response } from "express";
+import { z } from "zod";
+import { prisma } from "@/utils/prisma";
+
+const paramsSchema = z.object({
+  id: z.string().uuid(),
+});
+
+const sendAlreadyPublished = (res: Response) => {
+  res.status(409).json({
+    error: {
+      code: "ALREADY_PUBLISHED",
+      message:
+        "Agent templates with firstPublishedAt set cannot be hard-deleted",
+    },
+  });
+};
+
+export async function deleteHandler(req: Request, res: Response) {
+  const parsedParams = paramsSchema.safeParse(req.params);
+  if (!parsedParams.success) {
+    res.status(400).json({
+      error: "Invalid request params",
+      details: parsedParams.error.issues,
+    });
+    return;
+  }
+
+  try {
+    const template = await prisma.agentTemplate.findUnique({
+      where: { id: parsedParams.data.id },
+      select: { id: true, ownerAccountId: true, firstPublishedAt: true },
+    });
+
+    if (template === null) {
+      res.status(404).json({ error: "Agent template not found" });
+      return;
+    }
+
+    // Ownership guard: reject if caller is not the owner AND not an API key listener
+    const callerAccountId = res.locals.accountId;
+    const isApiKeyListener = res.locals.isApiKeyListener ?? false;
+    if (template.ownerAccountId !== callerAccountId && !isApiKeyListener) {
+      res.status(403).json({ error: "Not authorized to delete this template" });
+      return;
+    }
+
+    if (template.firstPublishedAt !== null) {
+      sendAlreadyPublished(res);
+      return;
+    }
+
+    // Atomic delete: a concurrent publish could set firstPublishedAt between
+    // the read above and the delete here, so re-check the invariant in the
+    // delete WHERE clause and return 409 if a publish slipped in.
+    const deleted = await prisma.agentTemplate.deleteMany({
+      where: { id: template.id, firstPublishedAt: null },
+    });
+
+    if (deleted.count === 0) {
+      const stillExists = await prisma.agentTemplate.findUnique({
+        where: { id: template.id },
+        select: { id: true },
+      });
+
+      if (stillExists !== null) {
+        sendAlreadyPublished(res);
+      } else {
+        res.status(404).json({ error: "Agent template not found" });
+      }
+      return;
+    }
+
+    res.status(200).json({
+      object: "agent_template",
+      id: template.id,
+      deleted: true,
+    });
+  } catch (error) {
+    req.log.error(
+      { error, stack: error instanceof Error ? error.stack : undefined },
+      "Failed to delete agent template",
+    );
+    res.status(500).json({ error: "Failed to delete agent template" });
+  }
+}
--- a/src/api/v2/agent-templates/handlers/delete.ts
+++ b/src/api/v2/agent-templates/handlers/delete.ts
@@ -31,6 +31,15 @@ export async function deleteHandler(req: Request, res: Response) {
     const callerAccountId = res.locals.accountId;
     const isApiKeyListener = res.locals.isApiKeyListener ?? false;
     if (template.ownerAccountId !== callerAccountId && !isApiKeyListener) {
+      req.log.warn(
+        {
+          callerAccountId,
+          templateId: template.id,
+          ownerAccountId: template.ownerAccountId,
+          action: "delete",
+        },
+        "Unauthorized agent-template access attempt",
+      );
       res.status(403).json({ error: "Not authorized to delete this template" });
       return;
     }
--- a/src/api/v2/agent-templates/handlers/delete.ts
+++ b/src/api/v2/agent-templates/handlers/delete.ts
@@ -6,16 +6,6 @@ const paramsSchema = z.object({
   id: z.string().uuid(),
 });
 
-const sendAlreadyPublished = (res: Response) => {
-  res.status(409).json({
-    error: {
-      code: "ALREADY_PUBLISHED",
-      message:
-        "Agent templates with firstPublishedAt set cannot be hard-deleted",
-    },
-  });
-};
-
 export async function deleteHandler(req: Request, res: Response) {
   const parsedParams = paramsSchema.safeParse(req.params);
   if (!parsedParams.success) {
@@ -29,7 +19,7 @@ export async function deleteHandler(req: Request, res: Response) {
   try {
     const template = await prisma.agentTemplate.findUnique({
       where: { id: parsedParams.data.id },
-      select: { id: true, ownerAccountId: true, firstPublishedAt: true },
+      select: { id: true, ownerAccountId: true },
     });
 
     if (template === null) {
@@ -45,29 +35,18 @@ export async function deleteHandler(req: Request, res: Response) {
       return;
     }
 
-    if (template.firstPublishedAt !== null) {
-      sendAlreadyPublished(res);
-      return;
-    }
-
-    // Atomic delete: a concurrent publish could set firstPublishedAt between
-    // the read above and the delete here, so re-check the invariant in the
-    // delete WHERE clause and return 409 if a publish slipped in.
+    // Hard delete is permitted in any publish state. Forks of this template
+    // have `forkedFromId` set to null (FK is ON DELETE SET NULL); generations
+    // (PR #204+) have `templateId` cleared the same way. The owner accepts
+    // that any cached/bookmarked URL pointing at this template will start
+    // 404'ing after delete.
     const deleted = await prisma.agentTemplate.deleteMany({
-      where: { id: template.id, firstPublishedAt: null },
+      where: { id: template.id },
     });
 
     if (deleted.count === 0) {
-      const stillExists = await prisma.agentTemplate.findUnique({
-        where: { id: template.id },
-        select: { id: true },
-      });
-
-      if (stillExists !== null) {
-        sendAlreadyPublished(res);
-      } else {
-        res.status(404).json({ error: "Agent template not found" });
-      }
+      // Lost a race with another concurrent delete.
+      res.status(404).json({ error: "Agent template not found" });
       return;
     }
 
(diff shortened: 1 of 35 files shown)


saulmc:
Thanks for the comprehensive review. Disposition of each item below — many were stale (the bot reviewed an earlier snapshot of the branch), a couple are wrong, a few were genuinely actionable.

## ✅ Addressed in this round (4 fixes)

- **#2 `publish.ts:61-75` race** — Same WHERE-pinned `updateMany` pattern that's already in `patch.ts`. First-publish only succeeds when `firstPublishedAt` is still null at write time; re-publish path is similarly pinned to `firstPublishedAt: { not: null }`. Both branches' `count === 0` short-circuits to 404. (`a93b4de`)
- **#5 `list.ts` `never_match` magic filter** — Replaced with an explicit early `200 { data: [], hasMore: false, nextCursor: null }` response. Same observable behavior, no Prisma query for invalid input. (`a710043`)
- **#10 `getSchemaBlock` to object args** — Matches the object-parameter convention from `.cursorrules` and the refactor I applied to the other test helpers earlier. (`a93b4de`)
- **#16 Concurrent slug-PATCH invariant** — Tightened from "sum == total" to also assert `successes >= 1` (guards against the genuine bug where all peers return 409). See "Wrong" below for why "exactly one wins" wasn't the right tightening. (`a93b4de`)

## 🪨 Already addressed before this review (stale snapshot)

The bot is comparing against an earlier branch state. All of these are fixed at HEAD:

- **#1 Migration FK ordering** — `20260508120812_authentication_api` → `20260508120813_add_agent_template_schema` → `20260508140000_seed_admin_account`. Correct order; verified via `ls prisma/migrations/`.
- **#3 detail.ts owner null check** — `detail.ts:97` is `if (expandOwner && owner === null)`.
- **#4 PATCH empty `prompt`** — `patch.ts:22-27` has `.refine((v) => v.trim().length > 0, …).optional()`.
- **#9 Import ordering** — fixed via `yarn lint --fix` across the listed files.
- **#11 (partial) Unused vars `tmplBare`/`tmplDraft`/`tmplUnlisted`/`tmplArchived`** — already dropped or unbound (`await createTemplate(…)` without assignment).
- **#17 FK delete_rule** — `schema.test.ts:151` asserts `expect(ownerForeignKey?.delete_rule).toBe("RESTRICT")`.
- **#18 Test infrastructure** — `auth-ownership.test.ts` and `auth-cross.test.ts` now wrap `app.listen` / `server.close` in awaited Promises; `cross.auth.test.ts` no longer collides on 4022 (port reallocation to 4070; metering moved to its rightful home on #200).

## ⚠️ Wrong / inaccurate

- **#11 (partial) `OTHER_ACCOUNT_ID` unused** — `grep` shows it used at `create.test.ts:161` (`ownerAccountId: OTHER_ACCOUNT_ID`) for a cross-account fixture. Not unused.
- **#16 "Exactly one wins" for concurrent slug PATCHes** — Tried this assertion (`expect(successes).toHaveLength(1)`); failed with `Received length: 2`. Each handler reads its own snapshot before issuing the WHERE-pinned write, so multiple PATCHes can each pin to a fresh state and succeed legitimately. The race fix in patch.ts prevents stale-state overwrites, not multiple updates against distinct snapshots. Kept the original sum-equals-total invariant and added `successes >= 1`.

## 🟡 Optional / dismissed on judgment

- **#6 Security logging on 403s** — Checked the repo's prevailing pattern. `auth-guards.ts:verifyDeviceOwnership`, `connections/{complete,delete}.ts`, and `agent-templates/{patch,delete,publish}.ts` all return 403 without `req.log.warn` before it. The two files that DO log (`generate-token.ts`, `notifications/subscribe.ts`) log a specific signal (Device disabled), not generic ownership failures. Adding it here would establish a new convention; better to apply uniformly as a separate change rather than only in this PR.
- **#7 `MAX_AUTO_SLUG_ATTEMPTS = 50`** — Slug uniqueness is scoped per `ownerAccountId` via `@@unique([ownerAccountId, slug])`. Exhausting 50 attempts requires a single owner to create 50+ templates from the same agent name. Not a real failure mode; explicit `slug` override is always available.
- **#8 List query WHERE complexity** — Real readability win, but extracting named WHERE-builder helpers is a non-functional refactor that should land as its own PR rather than nested inside review-comment cleanup.

## CodeRabbit comments (separately)

All 19 of CodeRabbit's inline comments on this PR are addressed and tagged `✅ Addressed` by its own automation — verified each against current code. Nothing else to action there.

saulmc:
Force-pushed (rebased on #195 + one new commit) — wired `buildUniqueSlug` into the create path so the row id is pre-picked to avoid cross-owner `<base>.<hash5>` URL collisions. Context for this is the [discussion on PR #195](https://github.com/xmtplabs/convos-backend/pull/195#discussion_r3222823856).

Specifically:
- `pickCollisionFreeId` (cross-owner predicate) lives in `create.ts`.
- `createWithExplicitSlug` and `createWithAutoSlug` both call it before insert.
- `createTemplateRow` now takes an explicit `id`.

Two follow-ups deliberately not in this PR:

1. **PR #204 (`feat/agent-template-generations`) needs the same wiring.** `services/generation-executor.ts:154` still calls `buildSlug(baseSlug, id)` directly with `randomUUID()`. While there: that file also stores `slug = "${baseSlug}.${slugHash(id)}"` in the DB, which appears inconsistent with the convention here (DB stores base slug only, URL form is reconstructed via `buildSlug(row.slug, row.id)`). Worth reconciling on #204 either way.

2. **Residual sub-ms read-then-write race.** `pickCollisionFreeId` reads existing rows, then `prisma.agentTemplate.create` inserts; a concurrent insert in that window could collide and bypass per-owner uniqueness. For that to actually mint a duplicate URL, two writes need to land sub-ms with the same base slug AND UUIDs that hash to the same 5 chars (~67M space). Negligible at our scale; SERIALIZABLE transaction or a stored `urlHash` cross-owner unique index can close it later without a separate migration if collisions ever surface.

tests/agent-templates.list.test.ts
```diff
@@ -0,0 +1,475 @@
...
+const readList = async (path = "/api/v2/agent-templates") => {
+  const response = await fetch(`${baseURL}${path}`, {
+    headers: await readerAuthHeaders(),
+  });
+  const body = (await response.json()) as ListEnvelope;
+
+  return { body, response };
+};
+
+const createTemplate = async (
```
neekolas, line 57: Think you can tell Claude to go through the tests and create some generic helpers. Seems like most of the test files have some version of this function.

src/api/v2/agent-templates/handlers/delete.ts
```diff
@@ -0,0 +1,65 @@
...
+      where: { id: parsedParams.data.id },
+      select: { id: true, ownerAccountId: true },
+    });
+
+    if (template === null) {
+      res.status(404).json({ error: "Agent template not found" });
+      return;
+    }
+
+    // Ownership guard: reject if caller is not the owner AND not an API key listener
```
neekolas, line 30: We should flag this `isApiKeyListener` thing as something we want to remove ASAP.

Having some magic API key that gives "god mode" powers to anyone who holds it is a pretty big back-door.

saulmc: Do you have a POV yet on the agent auth model that replaces this?

saulmc: It also is used by clients without user auth (web, twitter bot)

neekolas: I've been thinking about it a bit.   
  
High level: agents should be able to act as themselves, and their identity is always tied to an owner account ID.  
  
I'm still not sure the scope of what we want assistants to be able to do with the API.  
  
We can generate JWTs for each agent from the platform that links them to an account, and then expose that JWT to the assistant so they can use it freely.  
  
For web/twitter use-cases, we should treat them as unauthed. They can read things that are public. If we need them to write something, it would either have no owner or the owner would be us.

saulmc: I could see web/twitter being limited to requesting generations reading or listing templates. I don't know that we'd want those unauthed. Could make the case for reading single templates, maybe.

tests/agent-templates.list.test.ts
saulmc: ✅

neekolas (approved):

src/api/v2/agent-templates/handlers/delete.ts
neekolas: I mean, on the web you just are unauthed. You haven't signed in to anything, right?. We can rate limit the endpoints for unauthed users, but I think treating it that way is the most honest representation of reality.

Can definitely limit which endpoints are open to unauthed users to the bare minimum.

PR #197, opened 2026-05-08
feat: add template generator prompt and system-prompt module [STACKED ON #196]

saulmc:
## Template Generator + Builder SSE

**Re-stacked on PR #196** (CRUD on UUID Account model). Builder generate endpoint creates templates with UUID `ownerAccountId`.

### What this PR adds

| Route | Method | Auth | Response | Behavior |
|-------|--------|------|----------|----------|
| `/api/v2/agent_templates/generate` | POST | jwt or API key | JSON | Synchronous generation, returns template spec |
| `/api/v2/agent_templates/generate` | POST | jwt or API key | SSE | Streaming generation, `text/event-stream` |

### Architecture

```
POST /generate
  → system-prompt.ts (47KB verbatim prompt from pool)
  → templateGen.ts (OpenRouter integration, json_schema response_format)
  → create handler (persist as draft, UUID ownerAccountId = ADMIN_ACCOUNT_ID)
  → PostHog metering (builder.template.generated)
```

### Key components

| Component | Path | Purpose |
|-----------|------|---------|
| **System prompt** | `data/template-generator-prompt.txt` | 47,193 bytes, SHA256-verified verbatim copy from pool |
| **templateGen service** | `services/templateGen.ts` | OpenRouter with `json_schema` response_format, temp 0.7/0.2, BREVITY_RAIL |
| **Generate handler** | `handlers/generate-template.ts` | JSON + SSE modes, disconnect handling, keepalive |
| **PostHog metering** | `services/posthog.ts` | Lazy singleton, `builder.template.generated` event with metrics |

### Key decisions

- **UUID ownerAccountId**: Generated templates use `ADMIN_ACCOUNT_ID` as owner until SIWE auth lands.
- **SSE**: `text/event-stream` with `event: chunk` data frames, `event: done` sentinel, 15s keepalive pings.
- **OpenRouter**: `json_schema` response_format for structured output, with BREVITY_RAIL guard against overlong responses.
- **PostHog**: Lazy singleton, only captures when `POSTHOG_API_KEY` is set. Event includes `model`, `durationMs`, `inputTokens`, `outputTokens`.

### Stacking

```
fbac/authentication-api (PR #194)
  └─ feat/agent-templates-foundation (PR #195)
      └─ feat/agent-templates-crud (PR #196) ← this PR (#197)
```

<details>
<summary>Droid mission log</summary>

Mission: f9d21b7e-5b8b-4fcc-ae70-64eb803de10d
[...]

PR #200, opened 2026-05-09
feat: add template generator with per-account ownership [STACKED ON #199]

saulmc:
## Template Generator with Per-Account Ownership

Adds `POST /api/v2/agent-templates/generate` — synchronous LLM-backed template generation that persists results as draft `AgentTemplate` rows owned by the authenticated caller. Replaces the merged-too-early PR #197.

### What this PR adds

| **System prompt** | `data/template-generator-prompt.txt` | Verbatim copy of pool's skill-generator prompt |
| **`templateGen` service** | `services/templateGen.ts` | OpenRouter-backed generation (Claude Haiku), URL detection, GitHub passthrough, content classification, brevity-rail injection |
| **Generate handler** | `handlers/generate-template.ts` | JSON and SSE response modes; persists draft via `getEffectiveOwnerId(res)` |
| **PostHog metering** | `services/posthog.ts` | `builder.template.generated` event with `source`, `ownerAccountId`, `inputType`, token counts, latency |

### Flow

1. URL detection on input text
2. If URL → GitHub passthrough (repo + README extraction), Exa content extraction, or Twitter oEmbed (for x.com / twitter.com URLs)
3. Content-classifier passthrough where applicable
4. Main LLM call with strict JSON-schema `response_format`
5. Brevity rail appended to the production prompt
6. Persist as draft `AgentTemplate` owned by the authenticated account

### Reliability

- **15s timeout** on every outbound fetch (`AbortController` cleared in `finally`)
- **Lazy env reads** at call time, not import time — avoids test-time env races
- **t.co trust boundary documented**: redirects are delegated to Exa rather than fetched directly, so SSRF surface is bounded to Exa's protections
- **Soft defaults** for non-name fields; **server-injected** `connections: []`
- **PostHog fires on both success and error paths** so dead-letter analytics are visible
- **SSE mode** emits keep-alive comments every 15s; always responds 200 with a terminal `event: result` or `event: error` frame

### Body shape

PDF / image inputs are routed to a multimodal LLM call; text input is bounded by `MAX_TEXT_LEN`; base64 payloads bounded by `MAX_BASE64_LEN`. Comprehensive Zod validation surfaces 400 / 413 / 415 / 422 with `AppError`-typed responses.

### Auth model

`authOrAgentApiKeyAuth + requireAccount` (inherited from PR #199's posture). 401 for unauthenticated, 403 for JWT without `accountId`.

### Stacking

```
fbac/authentication-api (PR #194) ← PR #195 (foundation)
  └─ CRUD (PR #199)
      └─ this PR (builder)
          └─ create-job + twitter (PR #201)
```

### DIFF
--- a/data/template-generator-prompt.txt
+++ b/data/template-generator-prompt.txt
@@ -1,0 +1,363 @@
+You are a skill designer for Convos AI Assistants. **Group assistants are your specialty** — Convos is built around multi-party agents that live inside group chats, and you are an expert in the craft of designing them. You know what separates agents that thrive in a room full of humans (brevity, read-the-room instincts, group-level memory, multiplayer coordination moves, follow-through patterns) from ones that behave like 1:1 copilots and get annoying fast. Given any input — a one-sentence idea, a long-form description, a URL's extracted content, a PDF, an image, or pasted text — you generate a complete agent definition: the full system prompt, personality, name, and metadata. Infer the agent's purpose, audience, and personality from the source material.
+
+You follow the playbook below. Every agent you design is built from this DNA.
+
+---
+
+## Designing for Group Chat
+
+Convos is a group-chat-first platform. Solo or 1:1 framings are fine starting points — honor the creator's input. But your design job is to shape the agent so that as value accrues, the creator naturally wants to share it with friends (by adding the agent to other chats, or spinning up a multiplayer variant). Design choices that work in a group also work in a solo chat; choices that only work in a 1:1 dialogue break the moment a second person shows up. Default to the group case across every layer (BRAIN, SOUL, HEART, SCHEDULE):
+
+- **Most messages aren't for the agent.** In a room where humans talk to each other, silence is the default. Build trigger conditions into BRAIN so the agent speaks only when addressed, when its core job is triggered, or when it has a specific value-add. Agents that respond to every message kill the vibe.
+- **Every reply pings everyone.** Messages are push notifications on every member's phone; there is no private channel to one person. That's why brevity, reply threading (so members know who's being addressed), and "never leak one member's info to another" exist. Shape HEART and THE LINE around this.
+- **Memory is group memory.** Decisions, shared context, and observations belong to the room, not to individual users. Design the agent to track group-level dynamics — who tends to lead, who's quiet, emerging inside jokes, shared references — alongside per-user preferences.
+- **Multiple agents can coexist.** A chat may already have a trip planner, an expense splitter, and a sommelier. Stay in lane; don't drift into territory another agent (or a human member) is already handling.
+
+---
+
+## The Universal Agent Blueprint
+
+Every agent you design must cover these layers:
+
+### BRAIN — How It Thinks
+
+- **Primary Job** — The ONE thing this agent exists to do. Everything else is secondary.
+- **Decision Logic** — Rules it follows to make choices (e.g., weather rules for outdoor venues, budget thresholds, group consensus requirements).
+- **Memory & Tracking** — What it actively monitors in the conversation. Who said yes/no, running tallies, deadlines, preferences learned over time.
+- **Trigger Conditions** — When it activates vs. when it stays quiet. @mentions, keywords, time-based triggers, or context shifts.
+- **Proactive Behavior** — When it nudges the group without being asked. Deadlines approaching, missing responses, stalled conversations.
+
+### SOUL — Who It Is
+
+The personality that makes people WANT this agent in their chat.
+
+- **Character Name & Emoji** — A memorable identity (e.g., Open Claw 🦞, The Somm 🍷, Rally Bot 🏃).
+- **Personality Archetype** — One-line vibe (e.g., "the friend who's way too organized but roasts you for bailing").
+- **Tone** — Chill, witty, warm, dry, hype, calm. Pick a lane.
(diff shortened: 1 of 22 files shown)


saulmc:
**Triage response**

Going through all 8 items.

### 🔴 Critical

**#1 t.co SSRF.** Not "fixed" — residual concern acknowledged. `fb6fd33` dropped the direct `fetch(tco, { redirect: "follow" })` and routes t.co URLs through Exa, closing the SSRF surface **on our network**. The residual concern you flagged is real: a malicious tweet author can include a t.co link that redirects to an internal-to-Exa endpoint, and that content flows back into our generation. The blast radius now lives on Exa's infrastructure, not ours.

We accept that posture deliberately: Exa exists to fetch user-supplied URLs as its product, and our main `extractUrl(url)` flow (line 853) does the same thing with the user's own typed URL. The t.co path is just one more way to hand a URL to Exa. Made the trust boundary explicit in code at `5f01b9e`. Dropping the t.co loop entirely (the strictest fix) was the alternative I considered, but tweets are tiny — the meat is usually behind the link, and the LLM can see a t.co string but can't dereference it, so dropping the loop kneecaps the "build an agent from this tweet" flow.

**#2 Missing `await` on `.rejects.toThrow`.** Fixed in `fb6fd33` — `await` added to all five sites in `tests/template-gen.service.test.ts`. Also added `await-thenable` / `no-confusing-void-expression` to the eslint-disable header to match the existing convention in `template-gen.openrouter-errors.test.ts` (bun:test types `.rejects.toThrow` as `void`, so the runtime-correct pattern trips static rules without the disable).

### 🟠 Major

**#3 AppError migration.** Deferring. The handler already does the right thing (`error instanceof AppError ? error.statusCode : regex fallback`), so this is consistency, not behavior. And the actual repo idiom for service-layer code is plain `Error`, not `AppError`:

| File | AppError | Error |
|------|---------:|------:|
| `services/templateGen.ts` (this PR) | 3 | 14 |
| `services/provisioningClient.ts` | 0 | 8 |
| `services/job-executor.ts` | 0 | 1 |
| `handlers/invites/decode-invite-slug.ts` | 0 | 7 |
| `src/config.ts` | 0 | 8 |

`.cursorrules` does say "use AppError" — but no sibling service has been migrated. Doing it for just `templateGen.ts` here creates the same half-and-half state in one file that already exists across the directory. Proper scope: a follow-up PR that migrates `templateGen.ts`, `provisioningClient.ts`, and `job-executor.ts` together so we can drop the regex fallback in `errorStatus()`.

**#4 PostHog `logError` over `console.error`.** Same situation. `templateGen.ts` has five `console.error` sites (lines 172, 529, 743, 1034, etc.) and so do sibling services. Migrating only the posthog catch block is a wash. Better as part of the same logging-consistency follow-up.

**#5 Test duplicates production logic.** Pushing back — this is the established repo idiom, used by **six** guard test files:

```
tests/agent-templates.guard.test.ts
tests/agent-templates.read-guard.test.ts
tests/agent-templates.write-guard.test.ts
tests/agent-templates.generate.guard.test.ts
tests/agent-templates.generate.guard.sse.test.ts
tests/create-job.guard.test.ts
```

Every one defines a local `buildGuardedV2Router` and asserts `source.toContain('process.env.XMTP_ENV !== "production"')`. `agent-templates.guard.test.ts` even adds a *negative* `not.toMatch` against `import { XMTP_ENV } from "@/config"` to enforce the locked deviation that the env var is read raw.

The recommended fix (`import { createV2Router } from "@/api/v2/index"`) doesn't work as suggested — `src/api/v2/index.ts:42-48` exports a module-singleton, not a factory; the guard is evaluated once at import time, so flipping `process.env.XMTP_ENV` afterwards has no effect. The right fix is to refactor `src/api/v2/index.ts` to export `createV2Router()` and migrate all six guard tests together. That's a follow-up PR, not a PR #200 chore.

### 🟡 Minor

**#6 Tautological assertion.** Fixed in `fb6fd33`. Removed `startsWith("")` plus a duplicated `toContain("---")`.

**#7 Source-string match brittle.** Pushing back — this is the divergence detector for #5. If the local `buildGuardedV2Router` ever falls out of sync with the real guard in `src/api/v2/index.ts`, the source-string check is what catches it. Removing it weakens the test surface across all six guard test files. It stays until `createV2Router()` lands.

**#8 `system-prompt.ts` uses `console.error` instead of `logError`.** Fixed in `b9cbc53` — this one is at module init (import-time failure) rather than part of the broader `console.error`-throughout-the-feature concern.

### Also done as part of the triage (`b9cbc53`)

- `tests/template-gen.openrouter-errors.test.ts`: extracted `mockOpenRouterError(status, message)` helper; three 5xx/429/401 tests now one-liners.
- `tests/agent-templates.generate.disconnect.test.ts:245`: documented `process._getActiveHandles()` as a private Node/Bun API and pointed at `process.getActiveResourcesInfo()` as the modern replacement if the test ever breaks.

### Net

`fb6fd33` + `b9cbc53` + `5f01b9e`: 10 of 18 reviewer items resolved in code; #1's SSRF posture clarified with an in-code trust-boundary comment rather than claimed "fixed." 3 push-backs with reasoning (#5, #7, and CodeRabbit's multiline-import nit which is actually prettier-correct already). 4 nits deferred to a follow-up that bundles logging/AppError migration across the whole `services/` directory.

PR #205, opened 2026-05-12
feat: extend generations surface with twitter integration [STACKED ON #204]

saulmc:
## Summary

Adds optional `twitterContext` + ComposeReply stage to the async generation pipeline introduced in #204. **Replaces PR #201**.

## Changes

- `AgentTemplateGeneration` gains `twitterContext` (JSONB) and `reply` (TEXT) columns. Migration purely additive.
- `POST /agent-templates/generations` accepts optional `twitterContext { twitterHandle, tweetId, idea? }`. Validated via zod (handle regex `/^@?[A-Za-z0-9_]{1,15}$/`, numeric tweetId).
- Twitter intent moderation gate (`checkTwitterIntent`) runs after the universal content gate, only when `twitterContext` is present. Returns 422 `{ reason, category: "intent" }` on block.
- Generation executor extends with optional ComposeReply stage that runs only when `twitterContext` is present.
- `composeReply` has three fallback tiers: LLM → deterministic → minimal.
- `GET /generations/:id` response includes `reply: { text }` when populated.
- `compose-reply` service ported from the old `/create-job` line of work.
- `moderation` service extended with `checkTwitterIntent` (same OpenRouter Haiku model).

## Test plan

- [x] 135 agent-templates tests passing locally (9 new twitter cases on top of #204's 126)
- [x] Migration applied locally (`20260512000110_add_generation_twitter_fields`)
- [x] Typecheck clean
- [ ] CI green
- [ ] Migration applied to staging via deploy pipeline

## Stack

Stacked on #204 (which replaces #200). #199 → #204 → this PR.

## Replaces #201

This PR supersedes #201 entirely. The old `/create-job` surface (handlers, executor, provisioning client) never lands. Once approved + merged, please close #201.

## Client migration (vs. the deployed #201 surface)

| Client | Today's `/create-job` shape | This PR's `/generations` shape |
|---|---|---|
| iOS / web | `POST /create-job { source: "app", text, joinUrl }`, polls `/create-job/:id` | `POST /generations { source: "ios-app", inputs: { text } }` + `Idempotency-Key`, polls `/generations/:id`. iOS/web call their own provisioning client with the returned `templateId`. |
[...]

PR #207, opened 2026-05-13
feat: make agent-template read + generation endpoints public [STACKED ON #205]

saulmc:
## Summary

After discussion with @neekolas, the agent-template **read surface** (list + detail) and the **generation request/status endpoints** should be reachable without auth. Agents and the public template gallery both want to fetch templates and kick off generations without minting a token. Write endpoints stay locked down — only GET list, GET detail, POST /generations, and GET /generations/:id flip to optional auth.

Stacked on #205. Base will be retargeted to `otr-dev` once 199/204/205 land.

## What changes

**Middleware**
- Reintroduce `optionalAuthOrAgentApiKeyAuth`: skip auth when no credential headers are present, otherwise delegate to `authOrAgentApiKeyAuth`. Invalid creds still 401 — presenting them is opt-in.

**Routes** (router edits only — no schema changes)
- `GET  /agent-templates`              → optional auth
- `GET  /agent-templates/:idOrHashedSlug` → optional auth
- `POST /agent-templates/generations`    → optional auth
- `GET  /agent-templates/generations/:id` → optional auth
- All write endpoints (POST /, PATCH /:id, DELETE /:id, POST /:id/publish) unchanged — still `authOrAgentApiKeyAuth + requireAccount`.

**Handler semantics for anonymous callers**
- `listHandler`: no filter → published-only. `?status=draft|unlisted|archived` → 200 with empty data (no enumeration of non-public rows). `?status=published` → all published.
- `detailHandler`: published/unlisted/archived rows resolve normally; draft lookups 404 (the ownership branch fails because `undefined !== ownerAccountId`).
- `generationsPostHandler`: anonymous submissions default `ownerAccountId` to `ADMIN_ACCOUNT_ID`. Idempotency keying stays per `(ownerAccountId, idempotencyKey)`.
- `generationsGetHandler`: the generation UUID is treated as a bearer secret — anyone with the ID can poll status. Cross-account → 200 with the row.

**Tests**
- `auth-visibility.test.ts`: flip the 401 assertions to the new published-only / empty-result shape; add positive cases for anonymous list + anonymous published-detail.
- `agent-templates.read-guard.test.ts`: list route remains reachable in non-prod envs without auth.
- `agent-templates.generations-get.test.ts`: `cross-account access → 404` flipped to `cross-account → 200`, asserting capability-token semantics.

## No schema/migration changes

Confirmed: zero `prisma/migrations/*`, no `schema.prisma` edits, no SQL — purely handler/middleware/test code.

## Test plan
- [ ] `bun test tests/auth-visibility.test.ts` passes
- [ ] `bun test tests/agent-templates.read-guard.test.ts` passes
- [ ] `bun test tests/agent-templates.generations-get.test.ts` cross-account case asserts 200
- [ ] Manual: anonymous `curl /api/v2/agent-templates` returns only published rows
- [ ] Manual: anonymous `curl /api/v2/agent-templates?status=draft` returns `{ data: [], ... }`
- [ ] Manual: anonymous `curl -X POST /api/v2/agent-templates/generations` succeeds and the row is owned by `ADMIN_ACCOUNT_ID`

---
<a href="https://app.blacksmith.sh/xmtplabs/codesmith/convos-backend/pr/207"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need.</sup>

- [ ] Let Codesmith autofix CI failures and bot reviews

> [!NOTE]
> ### Make agent-template read and generation endpoints public with optional authentication
> - Adds a full CRUD + async generation API under `/api/v2/agent-templates` via a new [agent-templates.router.ts](https://github.com/ephemeraHQ/convos-backend/pull/207/files#diff-2faa4dc4ec7ffff7ca01b15362dd5ad613f5a8a7f932fbb54c2ec18738d348d9), including list, detail, create, patch, delete, publish, and generation start/poll endpoints.
> - Introduces `optionalAuthOrAgentApiKeyAuth` middleware so that list, detail, and generation endpoints are accessible without authentication while still enforcing auth when credentials are present.
> - Adds an async generation pipeline with idempotency, OpenRouter-backed template generation, content moderation, optional Twitter reply composition, SSE/long-poll response modes, and a TTL sweep service for stuck/expired jobs.
> - Adds new Prisma models `AgentTemplate` and `AgentTemplateGeneration` with supporting migrations, enums, and indexes.
> - Behavioral Change: importing `config.ts` now throws at startup if `GENERATION_STUCK_SWEEP_THRESHOLD_MS` ≤ `GENERATION_EXECUTOR_TIMEOUT_MS`; the TTL sweep also starts automatically on non-production server boot.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 75b407b. 26 files reviewed, 5 issues evaluated, 2 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> <details>
> <summary>src/api/v2/agent-templates/handlers/publish.ts — 0 comments posted, 1 evaluated, 1 filtered</summary>
>
> - [line 97](https://github.com/ephemeraHQ/convos-backend/blob/3d34e4c6f499ee137e6d55afbb77144ef709ae9f/src/api/v2/agent-templates/handlers/publish.ts#L97): The re-publish logic on lines 96-99 uses `template.status` which was fetched at line 42-44, but this value may be stale by the time we reach the re-publish branch. If another concurrent request (e.g., a PATCH) changes the template's status between the initial `findUnique` and the re-publish `updateMany`, the code will make decisions based on outdated data. For example: if the template was `"draft"` at fetch time but another request changed it to `"archived"` before the re-publish runs, `template.status === "draft"` would still be true, causing the code to set status to `"published"` instead of preserving `"archived"`. This contradicts the stated intent in the comment on lines 86-88 that "Status is normally preserved on re-publish." The check should be done atomically within the `updateMany` WHERE clause rather than using prefetched data. <b>[ Posting failed ]</b>
> </details>
>
> <details>
> <summary>src/api/v2/agent-templates/services/generation-executor.ts — 0 comments posted, 2 evaluated, 1 filtered</summary>
>
> - [line 144](https://github.com/ephemeraHQ/convos-backend/blob/3d34e4c6f499ee137e6d55afbb77144ef709ae9f/src/api/v2/agent-templates/services/generation-executor.ts#L144): `deriveBaseSlug` can return an empty string when `agentName` consists entirely of non-alphanumeric characters (e.g., emoji-only names like `"🤖"` or `"✨💫"`). After lowercasing, replacing non-alphanumeric chars with hyphens, and trimming leading/trailing hyphens, the result is `""`. This empty slug is then used in `persistTemplate` to create an `AgentTemplate` row with `slug: ""`, which could cause URL resolution issues or violate implicit non-empty expectations downstream. <b>[ Out of scope (triage) ]</b>
> </details>
>
>
> </details>
>
> 

## Release Notes

* **New Features**
  * Published agent templates are now publicly discoverable and readable without authentication
  * Generation status polling is accessible using only the generation ID, without requiring authentication
  * Anonymous users can submit generations to the system

* **Documentation**
  * Updated endpoint documentation to reflect public access behavior for published templates and capability-based polling

[Review Change Stack](https://app.coderabbit.ai/change-stack/xmtplabs/convos-backend/pull/207)

### DIFF
--- a/src/api/v2/agent-templates/agent-templates.router.ts
+++ b/src/api/v2/agent-templates/agent-templates.router.ts
@@ -28,17 +28,18 @@ agentTemplatesRouter.post(
 );
 
 // Async generation surface — must be mounted before /:idOrHashedSlug so the
-// wildcard doesn't capture "generations" as a slug-or-id.
+// wildcard doesn't capture "generations" as a slug-or-id. Both POST and GET
+// are public: anonymous submissions default to the ADMIN owner; the GET
+// status endpoint treats the generation ID itself as the capability token
+// (anyone with the UUID can poll).
 agentTemplatesRouter.post(
   "/generations",
-  authOrAgentApiKeyAuth,
-  requireAccount,
+  optionalAuthOrAgentApiKeyAuth,
   generationsPostHandler,
 );
 agentTemplatesRouter.get(
   "/generations/:generationId",
-  authOrAgentApiKeyAuth,
-  requireAccount,
+  optionalAuthOrAgentApiKeyAuth,
   generationsGetHandler,
 );
 
--- a/src/api/v2/agent-templates/agent-templates.router.ts
+++ b/src/api/v2/agent-templates/agent-templates.router.ts
@@ -1,5 +1,8 @@
 import { Router } from "express";
-import { authOrAgentApiKeyAuth } from "@/middleware/agentAuth";
+import {
+  authOrAgentApiKeyAuth,
+  optionalAuthOrAgentApiKeyAuth,
+} from "@/middleware/agentAuth";
 import { requireAccount } from "@/middleware/auth";
 import { createHandler } from "./handlers/create";
 import { deleteHandler } from "./handlers/delete";
@@ -12,12 +15,11 @@ import { publishHandler } from "./handlers/publish";
 
 export const agentTemplatesRouter = Router();
 
-agentTemplatesRouter.get(
-  "/",
-  authOrAgentApiKeyAuth,
-  requireAccount,
-  listHandler,
-);
+// Read endpoints (list, detail) are public. `optionalAuthOrAgentApiKeyAuth`
+// still sets res.locals.accountId when credentials are provided, so a
+// signed-in user can still see their own drafts; anonymous callers get
+// the published-only view.
+agentTemplatesRouter.get("/", optionalAuthOrAgentApiKeyAuth, listHandler);
 agentTemplatesRouter.post(
   "/",
   authOrAgentApiKeyAuth,
@@ -60,7 +62,6 @@ agentTemplatesRouter.post(
 );
 agentTemplatesRouter.get(
   "/:idOrHashedSlug",
-  authOrAgentApiKeyAuth,
-  requireAccount,
+  optionalAuthOrAgentApiKeyAuth,
   detailHandler,
 );
--- a/src/api/v2/agent-templates/handlers/detail.ts
+++ b/src/api/v2/agent-templates/handlers/detail.ts
@@ -58,12 +58,14 @@ export async function detailHandler(req: Request, res: Response) {
     });
 
     // If not found, check if it's a draft template accessible to the caller.
-    // accountId is guaranteed by the router (authOrAgentApiKeyAuth + requireAccount).
+    // The router uses `optionalAuthOrAgentApiKeyAuth`, so `accountId` may be
+    // undefined for anonymous callers — they never own a draft, so the
+    // ownership check below falls through and they get a 404.
     if (
       template === null &&
       uuidPattern.test(parsedParams.data.idOrHashedSlug)
     ) {
-      const accountId = res.locals.accountId as string;
+      const accountId = res.locals.accountId as string | undefined;
       const isApiKeyListener = res.locals.isApiKeyListener ?? false;
 
       const draftTemplate = await prisma.agentTemplate.findUnique({
@@ -71,8 +73,13 @@ export async function detailHandler(req: Request, res: Response) {
       });
 
       if (draftTemplate !== null && draftTemplate.status === "draft") {
-        // Draft is only visible to the owner or API key listener
-        if (draftTemplate.ownerAccountId === accountId || isApiKeyListener) {
+        // Draft is only visible to the owner or API key listener.
+        // `accountId === undefined` (anonymous) won't match the owner column.
+        if (
+          isApiKeyListener ||
+          (accountId !== undefined &&
+            draftTemplate.ownerAccountId === accountId)
+        ) {
           template = draftTemplate;
         }
         // If not authorized, template stays null → 404
--- a/src/api/v2/agent-templates/handlers/detail.ts
+++ b/src/api/v2/agent-templates/handlers/detail.ts
@@ -65,7 +65,7 @@ export async function detailHandler(req: Request, res: Response) {
       template === null &&
       uuidPattern.test(parsedParams.data.idOrHashedSlug)
     ) {
-      const accountId = res.locals.accountId as string | undefined;
+      const accountId: string | undefined = res.locals.accountId;
       const isApiKeyListener = res.locals.isApiKeyListener ?? false;
 
       const draftTemplate = await prisma.agentTemplate.findUnique({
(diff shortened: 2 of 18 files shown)


saulmc:
Thanks for the thorough review — running through each outstanding item with the current head of the branch (`d94c394`).

## 1. Idempotency namespace collision — addressed by UUID requirement; not a critical security risk

Status: addressed in commit `e696ab6` (Idempotency-Key must be a UUID).

Re-reading the review, I want to push back on the "🔴 Critical" rating and the "UUID enforcement doesn't prevent adversarial collisions" framing, because I don't think the attacks survive once UUID format is required.

| Attack named in the review | What it requires | Why it doesn't work post-UUID |
|---|---|---|
| **Generation-ID leak (anon → anon)** | Attacker submits with key `K`, then a legitimate caller picks the *same* `K` with the *same* body | The legitimate caller would have to independently generate the same 122-bit UUID. P ≈ 0 — that isn't an attack, it's a thermodynamic accident. |
| **Generation-ID leak (anon ↔ bot)** | Either side guesses the other's idempotency key | The agent API key is held by the bot and not exposed to outside callers; the bot's `idempotency-key` values are not observable, and UUIDs make them unguessable from outside. |
| **DoS / preemptive namespace occupation** | "Block legitimate submissions by occupying popular keys" | There are no "popular UUIDs" — a 122-bit namespace can't be camped. Each preemptive insert produces one row at attacker cost; it doesn't collide with anyone else's randomly-generated UUID. |

The residual case is "two anonymous callers happen to generate the same UUIDv4," which is mathematically negligible. Even if it landed, the only consequence is that they share a generation whose output is fully determined by inputs they both supplied — i.e. exactly the result either would have gotten independently. No PII leak, no privilege escalation, no DoS.

If you (or another reviewer) still want belt-and-suspenders here, I'd consider it a follow-up: e.g. "anonymous calls skip idempotency entirely." But I don't think the current state warrants a 🔴 Critical block on this PR.

## 2. Type assertion lint errors — already fixed

This was fixed in commit `9338bab` (titled `fix(lint): drop unnecessary string | undefined cast on res.locals.accountId`). The review pass that flagged it appears to have run against an earlier commit — `Typecheck, format check, lint` is green on `d94c394`.

## 3. twitterContext signed-nonce hardening — out of scope, deferring

Agreed the current `twitterContext` gate relies on the bot being the sole agent-API-key holder. I added the gate in commit `430094c` so anonymous + JWT-only callers can't attach `twitterContext` at all (the original threat was anonymous impersonation, which is closed).

A signed-nonce escalation (the bot signs `(twitterHandle, tweetId, ts)` with a separate key, backend verifies) is a meaningful defense-in-depth step if the agent API key ever leaks, but it's a meaningful piece of new infrastructure (key rotation, signature verification, bot-side signing service) and out of scope for this PR. Happy to file a follow-up ticket for it.

## 4. Generation-ID capability-token model — design is intentional; expiresAt already wired

Acknowledged. The capability-token model is deliberate and called out in the GET handler docstring. `expiresAt` is already part of the schema, the GET handler already 404s on expired rows, and `GENERATION_TTL_HOURS` controls the TTL sweep. So the "consider adding an `expiresAt` mechanism" recommendation is already in place.

User-facing documentation of the capability-token model is a fair callout; I'll add that to the OpenAPI YAML in a follow-up.

## 5. Test coverage for idempotency collision

Already covered:

- `tests/agent-templates.generations-post.test.ts > "non-UUID Idempotency-Key → 400"` (asserts non-UUID keys are rejected with the expected message).
- `tests/agent-templates.generations-post.test.ts > "same key + different body → 409"` (same body comparison is exercised against the dedupe path).
- `tests/agent-templates.generations-post.test.ts > "same key + same body → returns existing row"` (idempotency replay).

The specific anon-vs-anon collision case is academic with UUIDs (P ≈ 0), so I haven't added a test that fabricates a 122-bit coincidence.

---

TL;DR: items #1 and #2 don't block. #3 and the docs piece of #4 are reasonable follow-ups. CI is green on the head commit. Happy to discuss any of these further.

neekolas (approved):

PR #265, opened 2026-06-01
fix(builder): passthrough classifier leans agent-shaped, runs on minimax-m3

saulmc:
## Summary

Tunes the pre-generation **passthrough classifier** (`templateGen.classifyPastedContent`) — the step that decides whether pasted text is *already an agent prompt* (use verbatim) or *source material* (generate a persona from it).

- **Bias toward passthrough for agent-shaped input.** The prompt previously told the model to lean `false` when ambiguous and to require content be "ready-to-use as an agent prompt on its own." Reframed around "is this agent-shaped?", broadened the skill-definition shape (personas, role/identity lines, behavioral rules, more section headers, rough/partial definitions), and flipped the borderline tie-break to `true`. Genuine source material (articles, READMEs-for-humans, specs, excerpts) still returns `false`.
- **Run the classifier on a cheap, fast model.** New `BUILDER_CLASSIFIER_MODEL` env knob (default `minimax/minimax-m3`), replacing the main generation model (Opus) for this helper call. Concrete OpenRouter id so PostHog LLM Analytics prices its `$ai_generation`. Main generation and the GitHub selector are unchanged (still `getModel()`).

## Why

Agent-shaped pastes — a persona, a behavioral brief, a partial system prompt — were being classified as source material and **rewritten** by the generator instead of preserved verbatim, losing the author's wording. The stricter-than-needed prompt and the over-powered model were both contributing; this leans the call toward respecting a real prompt when it sees one.

## Notes

- Behavior unchanged below `PASSTHROUGH_MIN_LENGTH = 300` and beyond the 8k-char classify window — those floors still bound what's even considered.
- `BUILDER_CLASSIFIER_MODEL` is overridable per-env if `minimax/minimax-m3` needs swapping.

---
<a href="https://app.blacksmith.sh/xmtplabs/codesmith/convos-backend/pr/265"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782933519&installation_id=128169237&pr_number=265&repository=xmtplabs%2Fconvos-backend&return_to=https%3A%2F%2Fgithub.com%2Fxmtplabs%2Fconvos-backend%2Fpull%2F265&signature=ef41735519e84d0e157dccda135a2f7937e576802d52ea2673e7a9efdbd39184"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need. Autofix is disabled.</sup>

> [!NOTE]
> ### Fix passthrough classifier to lean agent-shaped and run on `minimax-m3`
> - Adds `BUILDER_CLASSIFIER_MODEL` to [config.ts](https://github.com/ephemeraHQ/convos-backend/pull/265/files#diff-c3095d5010e65c52737a98a5d618ea24049ebe90c8470752426081d70ed6e012), defaulting to `minimax/minimax-m3`, so the passthrough classifier uses a dedicated model separate from the main builder model.
> - Updates `classifyPastedContent` in [templateGen.ts](https://github.com/ephemeraHQ/convos-backend/pull/265/files#diff-d151881c4ab5cd2005df4016ac6a564fb6094e4ce06b69fbf9529fb70e59148a) to call `getClassifierModel()` instead of `getModel()`, and updates the prompt to bias toward classifying agent-shaped content as passthrough.
> - Adds a test hook `__setClassifierModelOverrideForTests` to allow overriding the classifier model in tests.
>
> 
> #### Changes since #265 opened
>
> - Added model tracking to passthrough operations and updated metrics reporting [f6b772d]
> - Added test coverage for classifier model metrics reporting during content passthrough [f6b772d]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized de22bbe.</sup>
> 
>
> 

## Summary by CodeRabbit

* **Refactor**
  * Improved pasted content classification in agent template generation with refined decision logic.
  * Separated classifier model configuration from the main generation model to optimize performance and accuracy.

### DIFF
--- a/src/api/v2/agent-templates/services/templateGen.ts
+++ b/src/api/v2/agent-templates/services/templateGen.ts
@@ -10,8 +10,9 @@
  *
  * OpenRouter raw fetch to https://openrouter.ai/api/v1/chat/completions with
  * strict response_format json_schema, temp 0.7, no max_tokens.
- * Helper calls (GitHub-instructions selector, content-classifier) at temp 0.2
- * without response_format.
+ * Helper calls run at temp 0.2 without response_format: the GitHub-instructions
+ * selector uses the main model; the content-classifier uses the cheap
+ * BUILDER_CLASSIFIER_MODEL and leans toward passthrough.
  *
  * Soft defaults for non-name fields. Server-injects connections: [].
  */
@@ -20,6 +21,7 @@
 
 import { APIConnectionTimeoutError, APIError, APIUserAbortError } from "openai";
 import {
+  BUILDER_CLASSIFIER_MODEL,
   BUILDER_EXA_SERVICE_KEY,
   BUILDER_MODEL,
   BUILDER_OPENROUTER_API_KEY,
@@ -91,6 +93,7 @@ function describeLlmError(err: unknown): string {
 
 let _apiKeyOverride: string | null | undefined = undefined;
 let _builderModelOverride: string | null = null;
+let _classifierModelOverride: string | null = null;
 let _exaKeyOverride: string | null | undefined = undefined;
 let _systemPromptOverride: string | null = null;
 
@@ -122,6 +125,19 @@ export function __setBuilderApiKeyOverrideForTests(
 /** Override `BUILDER_MODEL` for tests. Pass `null` to clear. */
 export function __setBuilderModelOverrideForTests(model: string | null): void {
   _builderModelOverride = model;
+}
+
+/** Model for the passthrough classifier (`classifyPastedContent`) — a cheap
+ *  model separate from the main `getModel()`. Override `BUILDER_CLASSIFIER_MODEL`. */
+export function getClassifierModel(): string {
+  return _classifierModelOverride ?? BUILDER_CLASSIFIER_MODEL;
+}
+
+/** Override `BUILDER_CLASSIFIER_MODEL` for tests. Pass `null` to clear. */
+export function __setClassifierModelOverrideForTests(
+  model: string | null,
+): void {
+  _classifierModelOverride = model;
 }
 
 /** The system prompt actually used for generation — the loaded file unless an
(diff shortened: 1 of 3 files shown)


neekolas (approved):

neekolas (approved):

saulmc:
Thanks — addressed in f6b772d.

**🔴 #1 Passthrough metrics model attribution — fixed (with a correction).** You're right that content-passthrough reported `getModel()` for tokens that came from the minimax classifier. But the fix isn't a blanket `getClassifierModel()` at both return sites — the GitHub-passthrough path is more subtle:

- GitHub **repo scan** → tokens come from the *selector*, which runs on `getModel()` (Opus). `getModel()` was already correct there.
- GitHub **direct file link** → `tryGithubPassthrough` delegates to `tryContentPassthrough`, so those tokens come from the *classifier* (minimax) — `getModel()` would be wrong, same as the content path.

So a single model isn't right for the GitHub return site. Instead, `PassthroughBundle` now carries the model that produced its tokens (selector path sets `getModel()`, classifier/direct-file paths set `getClassifierModel()`), and both return sites read `bundle.model`. This attributes all three origins correctly.

**🧪 #3 Test coverage — added.** New test `content passthrough metrics report the classifier model + its tokens` asserts `metrics.model === BUILDER_CLASSIFIER_MODEL` on the passthrough path (would have caught #1).

**🟡 #2 Function signatures — keeping as-is.** Explicit `: string` return type and positional params match the existing sibling seams (`getModel`, `__setBuilderModelOverrideForTests`) in this file; switching to inferred returns / object params would make the new functions inconsistent with their neighbors. (Noted you flagged this as optional given the precedent.)

**🔵 #4 `getClassifierModel()` unit test — skipping.** It's a one-line `?? config` accessor with no logic of its own, and it's now exercised end-to-end by the passthrough-metrics test above; a dedicated unit test would just restate the implementation.

PR #275, opened 2026-06-02
feat(credits): wire signup bonus on first account creation

fbac:
## Summary

Wires the previously-dead `signup_bonus` GrantKind so it is granted exactly once when a new `accountId` is first created (the SIWE upgrade on `POST /v2/auth/token`).

- **Grant kind**: corrected the stale seed description ("first agent creation" — never wired) via a data migration. `id` (`signup_bonus`) and `name` unchanged.
- **Config**: new env var `PAYMENTS_SIGNUP_BONUS_CREDITS`. Unlike the other `PAYMENTS_*` vars, this one is **optional**: `0`/unset disables the grant (kill-switch). Ships OFF until ops sets a value.
- **Atomic grant**: the bonus ledger write happens **inside the same transaction** as `account.create` + `authMethod.create`, via an injectable `onCreate` hook on `upsertAuthMethodAndAccount` (+ a `grantSignupBonusWithTx` helper). Account + bonus commit or roll back together.
- **Idempotency**: key `signup_bonus_<accountId>` (matches the canonical `idempotencyKeySchema` charset `^[A-Za-z0-9_-]{1,255}$` — no `:`), with a test asserting conformance.

### Durability (addresses review feedback)
The earlier best-effort approach gated the grant on `created` and swallowed failures — a transient grant failure (or a process death between the account-create commit and the grant) would leave `created === false` forever, permanently suppressing the bonus.

The atomic approach removes that window entirely **without** a tracking column or a per-login check:
- A new account can never exist without its bonus — they share one transaction.
- A transient ledger failure rolls the account back and surfaces as a **retryable 500**, not a silent loss. The next login retries and self-heals.
- Only touches the rare first-SIWE path; zero lock contention (fresh account), zero cost on repeat logins.

### Invariant (per-account, not per-person)
A user who loses their wallet and signs in with a new wallet gets a new account + a fresh bonus — intended. A **future** account-recovery flow (Apple/Google subscription key) MUST reuse the existing `accountId`; reusing it yields `created === false`, suppressing a second bonus. Per-person dedup is a separate future slice.

### Follow-up (not in this PR)
- Sibling Terraform PR adds `PAYMENTS_SIGNUP_BONUS_CREDITS` to the ECS task definition. Dormant in prod until applied.
- Separate PR: enforce the idempotency-key charset on internal keys + migrate the deployed `daily_refill:` key format to `_` (forward-only; the daily rate-limit keys off `grantKindId`+date, not the key).

## Test Plan
- [x] `pnpm typecheck` — clean
- [x] `pnpm lint` — 0 errors on touched files
- [x] `pnpm format:check` — clean
- [x] `pnpm test` — 806 passed, 1 skipped (1 unrelated pre-existing wall-clock flake in agent-templates executor, passes on isolated re-run)
- Coverage: config loader; repo `created` flag + race winner; **atomic rollback + throw-once-then-heal (self-heal)**; handler integration (new account → one `signup_bonus` grant of configured amount + key-charset assert; relogin → no second bonus)

> [!NOTE]
> ### Grant signup bonus credits on first SIWE account creation
> - On first-time SIWE login, `grantSignupBonusWithTx` is called atomically within the account creation transaction, crediting `signupBonusCredits` to the new account with idempotency key `signup_bonus_{accountId}`.
> - `upsertAuthMethodAndAccount` in [repository.ts](https://github.com/ephemeraHQ/convos-backend/pull/275/files#diff-0ff968ca7965fe91033bcae8ae9ed9d246368ecd4821897c03c72f0e39427b4a) now returns `{ accountId, created }` and accepts an optional `onCreate` callback executed inside the same transaction.
> - Signup bonus amount is configured via `PAYMENTS_SIGNUP_BONUS_CREDITS` env var (defaults to 0); invalid values throw at config load time.
> - Risk: if the bonus grant fails, the entire account creation rolls back and the request returns HTTP 500.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8ae0420.</sup>
> 
>

### DIFF
--- a/prisma/migrations/20260602120000_signup_bonus_account_creation/migration.sql
+++ b/prisma/migrations/20260602120000_signup_bonus_account_creation/migration.sql
@@ -1,0 +1,7 @@
+-- Correct the signup_bonus grant-kind description. The original seed
+-- (20260515120000_payments_credits_foundation) said "Granted once on first
+-- agent creation", which was never wired. The bonus is granted once on first
+-- account creation (SIWE upgrade on POST /v2/auth/token).
+UPDATE "GrantKind"
+SET "description" = 'Granted once when an account is first created (SIWE upgrade)'
+WHERE "id" = 'signup_bonus';
--- a/src/accounts/repository.ts
+++ b/src/accounts/repository.ts
@@ -5,7 +5,7 @@ import { prisma } from "@/utils/prisma";
 export async function upsertAuthMethodAndAccount(args: {
   type: AuthMethodType;
   externalKey: string;
-}): Promise<{ accountId: string }> {
+}): Promise<{ accountId: string; created: boolean }> {
   const findOrInsert = () =>
     prisma.$transaction(async (tx) => {
       const existing = await tx.authMethod.findUnique({
@@ -13,7 +13,7 @@ export async function upsertAuthMethodAndAccount(args: {
           type_externalKey: { type: args.type, externalKey: args.externalKey },
         },
       });
-      if (existing) return { accountId: existing.accountId };
+      if (existing) return { accountId: existing.accountId, created: false };
 
       const account = await tx.account.create({ data: {} });
       await tx.authMethod.create({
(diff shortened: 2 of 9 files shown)


lourou (approved):
LGTM!

neekolas (approved):

PR #309, opened 2026-06-15
feat(agent-templates): server-side distill — preview + progressPhrases on the generation poll

saulmc:
Moves the agent-builder **distill** step into the backend so the generation poll narrates the build and pre-fills the draft agent. The app/skill becomes a thin poller — this is what makes the in-chat `agent-builder` skill deletable (CON-527).

## Poll contract

`GET /api/v2/agent-templates/generations/{id}` (and POST's long-poll replies):

```
202 (running)   { status:"running", progressPhrases:[…], preview:{ agentName, emoji, description } }
202 (pending)   { status:"pending" }                       // before distill writes
200 (done)      { status:"done",  templateId }             // client fetches the template
200 (failed)    { status:"failed", error }
```

`preview` (the draft agent) and `progressPhrases` ride **only the in-progress 202s** and are dropped on the 200 — by then `templateId` points at the real template, so there's nothing to "complete." GET previously always returned 200; POST already did 202/200.

Backward-compatible: every existing caller (skill `build.mjs`, website `start`/`status`, twitter-bot) guards on `res.ok`, and `202 ∈ 2xx`.

## What changed

- **`services/distill.ts`** — one fast OpenRouter call (strict json_schema, builder model) producing identity + `progressPhrases`, a faithful port of the skill's distill recipe (`data/distill-prompt.txt`). Best-effort: a failure logs and the build still completes. Skipped for the twitter path (no progress UI) and text-less inputs.
- **Executor** — runs distill after the atomic claim / before generate, writes `preview` + `progressPhrases` in one `status='running'`-guarded update, and feeds the distilled identity (caller-pinned `prefill` wins per field) into generate + persist so the early card matches the final template. `markDone` is unchanged.
- **Handlers** — `GET` returns `202` while in progress / `200` when terminal; both handlers surface `preview`/`progressPhrases` only on non-terminal rows (shared `lib/generation-preview.ts`).
- **Schema** — two nullable `jsonb` columns (`preview`, `progressPhrases`) on `AgentTemplateGeneration` + migration. Idempotency, moderation, and publish-status behavior unchanged.
- Refactored the prompt loader into a shared `lib/data-prompt.ts` (`loadDataPrompt`) so distill and the system prompt share the bundled-vs-source path resolution.

## Testing

- `pnpm check` (typecheck + format + lint): green.
[...]

PR #310, opened 2026-06-16
feat(agent-templates): multi-attachment generation inputs via presigned upload (drop base64)

saulmc:
Stacked on #309. Moves the agent-builder's binary inputs **off the request body**: the app uploads each attachment to a **private** S3 bucket via presigned PUT, and the generation request carries lightweight **object-key references** instead of base64. The backend reads the bytes itself for generation + moderation, so decrypted conversation content never gets a public CDN URL. Enables **text + N attachments** (the Convos multi-attachment message) — image, PDF, and voice.

## Wire change (breaking)

`inputs.attachments: [{ objectKey, mimeType, filename? }]` (array, count cap 9) **replaces** the `pdfBase64` / `imageBase64` scalars. Clients (iOS + web) must adopt the presigned-upload flow before this lands. Attachments live in the existing `inputs` JSON column — **no migration**.

```
GET  /api/v2/agent-templates/attachments/presigned?contentType=…  → { objectKey, uploadUrl }   (private bucket, no public asset URL)
POST /api/v2/agent-templates/generations  inputs.attachments: [{ objectKey, mimeType, filename? }]
```

## What changed

- **`services/build-attachments.ts`** — private-bucket client: presign-PUT, `headBuildObject`, `getBuildObjectBytes` (the repo's first `GetObject`), the MIME allowlist + class-specific per-file size caps, and `build/`-prefix fencing on every read (a caller can't point the generator at an arbitrary object in a shared private bucket).
- **`services/attachment-resolver.ts`** — shared resolve stage: fetch bytes → moderate (images) / transcribe + moderate (audio) → build LLM content blocks. Used by both the async executor (moderate) and the admin ephemeral endpoint (trusted, no moderation).
- **Executor** — resolves attachments off the request path, builds an **N-block multimodal** message (one `image_url` / `file` per attachment), and **folds voice transcripts into the text directive** so distill + the text path see them and the generator only handles images/PDFs.
- **Submit-time fail-fast caps** — count / unknown type / per-file + aggregate size / unfetchable reference → **4xx**, before the executor runs.
- **`services/image-moderation.ts`** — AWS Rekognition `DetectModerationLabels` by S3 reference, fail-open like the text moderator. Images are restricted to **PNG/JPEG** (the formats Rekognition reads) so every accepted image is moderatable.
- **`services/transcribe.ts`** — voice → text via OpenRouter `input_audio` on `BUILD_TRANSCRIBE_MODEL`, behind a single seam so the provider is swappable.
- **Schema** — two nullable-aware refinements to the poll preview shaping; `inputs.attachments` rides the existing JSON column. **No DB migration.**
- **Config** — `PRIVATE_ASSETS_BUCKET`, `BUILD_ATTACHMENTS_MAX_COUNT` (9), `BUILD_ATTACHMENTS_MAX_TOTAL_BYTES`, `BUILD_TRANSCRIBE_MODEL`; adds `@aws-sdk/client-rekognition`.

## Deliberate decisions (worth a look in review)

- **Binary moderation is async → terminal `failed`, not a submit-time 422.** Rekognition over N images + transcription are too slow to gate the POST. Text moderation stays synchronous (422).
- **Images reach the LLM as server-fetched data URIs, not presigned-GET** — honors "the backend pulls the bytes itself, never a public URL."

## Testing

- `pnpm check` (typecheck + format + lint): green.
- Full suite: **1061 passed, 1 pre-existing skip**. New coverage: `build-attachments`, `image-moderation`, `transcribe`, `attachment-resolver`, the presigned endpoint, submit-time caps (count/type/size/unfetchable → 4xx), executor resolve → N-block + Rekognition-block → `failed` + audio fold.
- **Not done:** a live audio-format transcription run (needs `BUILDER_OPENROUTER_API_KEY`) — see the risk below.

## Known risk

**Audio format.** OpenRouter `input_audio` is reliable for wav/mp3; mobile records m4a/aac/ogg. The provider sits behind one seam, but the actual m4a path needs a live smoke test — if the model rejects the container it's a server-side transcode or an STT-provider swap.

## Follow-ups (out of this PR)

- Provision `PRIVATE_ASSETS_BUCKET` + a short lifecycle-expiry rule; add `s3:GetObject` + Rekognition `DetectModerationLabels` to the IAM role.
- iOS + web ship the presigned-upload flow before the base64 removal lands.

https://linear.app/convos/issue/CON-533

---
<a href="https://app.blacksmith.sh/xmtplabs/codesmith/convos-backend/pr/310"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784163640&installation_id=128169237&pr_number=310&repository=xmtplabs%2Fconvos-backend&return_to=https%3A%2F%2Fgithub.com%2Fxmtplabs%2Fconvos-backend%2Fpull%2F310&signature=68f989aea4de4335141f70c427e9cfab9d5784307e675d0ff8728e18e97ca2c4"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

> [!NOTE]
> ### Replace base64 attachment inputs with presigned S3 upload references in agent template generation
> - Adds a `GET /api/v2/agent-templates/attachments/presigned` endpoint that issues presigned S3 PUT URLs for uploading attachments to a private bucket, rate-limited to 20 req/min per IP.
> - Replaces `pdfBase64`/`imageBase64`/`mimeType`/`filename` fields in generation POST bodies with an `attachments` array of `{ objectKey, mimeType, filename? }` references; both `POST /generations` and `POST /generations/ephemeral` are updated.
> - Adds a new [attachment-resolver.ts](https://github.com/ephemeraHQ/convos-backend/pull/310/files#diff-4b3d4ea46d7a5a6dbeb246d35ae9a76c8e94ff2af11a8b66dad4d587c7b0eda3) service that fetches attachment bytes from S3, enforces per-kind and aggregate size caps, transcribes audio via OpenRouter, and moderates images via AWS Rekognition (fail-open on errors).
> - Adds [build-attachments.ts](https://github.com/ephemeraHQ/convos-backend/pull/310/files#diff-012527fa6cb28f27c2e51485c3ea0a670462a291180ad2d2701e7e318ce9da44) for S3 interactions: presigning, HEAD validation, and byte retrieval with MIME allowlist and per-kind size limits.
> - Updates [distill.ts](https://github.com/ephemeraHQ/convos-backend/pull/310/files#diff-275f58f94ccc137bf59443d651f7a49d89e020f0be585ee9b3bed4da36e785a1) and [templateGen.ts](https://github.com/ephemeraHQ/convos-backend/pull/310/files#diff-d151881c4ab5cd2005df4016ac6a564fb6094e4ce06b69fbf9529fb70e59148a) to accept `attachments: ResolvedAttachment[]` with multimodal `image_url` and `file` content blocks instead of raw base64.
> - Non-terminal generation responses now include `estimatedDurationMs` (higher when attachments are present) from new config values `GENERATION_ESTIMATE_MS` and `GENERATION_ESTIMATE_WITH_ATTACHMENTS_MS`.
> - Risk: existing clients sending `imageBase64`/`pdfBase64` fields will receive 400 errors; callers must upload files to S3 first and pass object keys instead.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 50fd77e.</sup>
> 
>

## New Features
- Added attachment-based generation inputs (images, PDFs, and audio) with server-side resolution, transcript handling, and moderation-aware failures.
- Introduced `GET /api/v2/agent-templates/attachments/presigned` to fetch presigned upload details for generation attachments, with rate limiting.
- Added Rekognition-based image moderation (with configurable minimum confidence) and updated moderation behavior/tuning via environment settings.
- Added `estimatedDurationMs` to in-progress generation status responses, accounting for attachment overhead.
- Added audio transcription support for attachment-based workflows.
- Updated environment documentation and configuration for private asset storage, size caps, and moderation/transcription thresholds.


src/api/v2/agent-templates/services/build-attachments.ts
```diff
@@ -0,0 +1,217 @@
...
+
+export interface PresignedBuildUpload {
+  objectKey: string;
+  uploadUrl: string;
+}
+
+/** Mint a presigned PUT for a new build attachment. The returned `objectKey` is
+ *  what the client echoes back in `inputs.attachments[]`. No public asset URL is
+ *  returned — the bucket is private and only the backend reads it. */
+export async function presignBuildUpload(
```
lourou, line 151: My first instinct was to reuse the existing public bucket and just mark these files private, we'd inherit the 30-day expiry and existing file infra for free. I dug in and revised that: our public bucket is public at the bucket-policy + CloudFront layer (the `PublicReadGetObject` statement grants `s3:GetObject` to `*` on `/*`, CloudFront fronts the whole bucket, and per-object `public-read` ACLs were dropped in `0588d59`), so a per-object "private" flag wouldn't protect these decrypted-conversation attachments, anything in that bucket is reachable at `CDN_BASE_URL/<key>`. So a separate private bucket is the right call.

Worth speccing it explicitly as a private, transient agent-ingest dropbox:
- presigned **PUT only** write/create-only, no client delete/list/get, and no overwrite (so a validated object can't be swapped afterwards)
- readable **only by our own agent task role** via IAM
- a **short 24h lifecycle expiry** doing all the cleanup (no app-side deletes)

src/api/v2/agent-templates/services/build-attachments.ts
```diff
@@ -0,0 +1,217 @@
...
+    throw new AppError(400, `Unsupported attachment type: ${contentType}`);
+  }
+  const ext = mime.extension(contentType);
+  const objectKey = `build/${uuidv4()}${ext ? `.${ext}` : ""}`;
+  const command = new PutObjectCommand({
+    Bucket: bucket,
+    Key: objectKey,
+    ContentType: contentType,
+  });
+  const uploadUrl = await getSignedUrl(client, command, { expiresIn: 3600 });
```
lourou: The presigned PUT here is signed with only Bucket/Key/ContentType, no size condition, so the URL authorizes an upload of any size for the full hour. The per-file/aggregate caps only run later at generation submit, not on the upload itself, so someone could mint presigns and PUT arbitrarily large objects without ever submitting (storage-stuffing).

Worth adding a ceiling: `createPresignedPost` with a `content-length-range` condition so S3 rejects oversize uploads at PUT time. It's a small wire change (multipart POST instead of raw PUT).

saulmc: Kept the PUT but sign content-length into the URL (and check it against the per-kind cap before minting), so S3 rejects any body that doesn't match the exact signed size — declaring small and pushing big breaks the signature. Same size ceiling as content-length-range, minus the multipart POST change.

For the orphan side (minting presigns and never submitting), the expiry rule on the build/ prefix will age them out, and there's a 20/min per-IP limit.

saulmc:
Thanks for the thorough review. Walking each item against the current branch — several were already addressed in commits that landed after this review ran (including the #314 hardening pass):

### Already addressed
- **#2 `headBuildObject` ContentLength** (critical) — now throws `AppError(400, "Attachment size unavailable")` when `ContentLength` is missing, exactly as recommended (#314).
- **#4 Attachment order in multimodal generation** — already preserved; the builder maps over the original `attachments` array, not the filtered image/PDF lists (the quoted snippet is stale — see the `// Map the original attachments array … so a mixed image/PDF order from the caller is preserved` comment).
- **#5 `BUILD_ATTACHMENTS_MAX_COUNT` hard cap** — `Math.min(9, …)` in config.
- **#6 Twitter `idea` whitespace** — normalized to "absent when whitespace-only" before the `??` fallback, so a valid `inputs.text` still satisfies the intent check.
- **#7 Ephemeral timeout** — returns `504 Generation timed out` on the timeout signal, distinguished from client-disconnect.
- **#3 Rate limiter** — present (`buildAttachmentPresignedLimiter`), as you noted.
- **Minor** — attachment-resolver now throws `AppError` for unsupported/over-cap; `getBuildObjectBytes` NotFound + no-Body tests added; presigned-test port deconflicted.

### Deliberate — deferring to CON-534
- **#1 Image moderation fail-open** — intentional parity with the already-shipped text `moderation.ts`, which is also fail-open. The fail-closed-on-misconfig nuance is reasonable, but applying it only to images would diverge from text moderation, and the broader ingest-time content-scan work (including a fail-closed posture) is scoped to **CON-534**. Leaving as-is here for consistency.
- **Live audio transcription** — tracked as a Known Risk in the PR description.

### Optional, not blocking
- Cache-header assertions in `build-attachment-presigned.test.ts` — behavior is correct, just not asserted.
- Object-parameter signatures (`presignBuildUpload`, `transcribeAudio`) — style preference; keeping signatures stable mid-PR.

Net: the Must-Fix / Should-Fix items are already in the branch; **#1** is the one conscious parity decision, pushed to CON-534.

PR #324, opened 2026-06-23
feat(credits): single materialized wallet ledger for subscriptions

lourou:
## Problem
Subscribers were reading ~0 credits. Root cause: the subscription balance was **derived/phantom** (computed live = `tierGrant − consumes`, gated on a live entitlement check) and the read path was **bimodal** — so any entitlement drift (stale `currentPeriodEnd`, missed renewal) dropped a sub from the derived bucket to the raw bucket ≈ 0. The two-bucket model was the underlying source.

## Approach
Collapse to **one materialized wallet ledger** (`UserCredits.balance` + `CreditLedger`) as the only spendable truth. Subscriptions become money-in / money-out on that wallet:

- **Grant** on subscribe/renew — a real `sub_grant` ledger row, **idempotent** per `{subscription, periodStart}` (dedups across the verify + renewal-notification paths).
- **Forfeit** on expiry/refund/revoke — `−min(walletBalance, max(0, periodGrant − periodConsumes))`, computed **under a row lock** + `floorCheck` so it can never go negative or wipe admin/promo/signup credits. Cancel-while-active is a no-op.
- **Read** collapses to `getBalance` (no derivation, no bimodal switch); **consume** is one path for everyone (`recordOnly` removed). iOS `/credits` JSON is byte-identical.
- **Tiers → PLUS-only**: dropped the vestigial `PAYMENTS_GRANT_BUILDER_MONTHLY` env fallback. The legacy `builder|pro` *SKU* regex stays — grandfathered receipts still carry those SKUs and map to Plus.
- Boot-validate `PAYMENTS_GRANT_PLUS_MONTHLY`.

Deliberately **n=1-scoped** (one subscriber): no cutover state machine / dual-write / backfill harness. A one-time idempotent materialization script (`dev/scripts/materializeSubscriptionWallets.ts`, dry-run by default) seeds the existing sub's current period; otherwise it re-grants on next renewal.

## Tests
`tests/{subscriptions,payments,credits,credits-admin}` → **331 pass** (the 4 red suites are a pre-existing `googleapis`-absent module-load gap; the Google path is mock-tested locally, real in CI). Includes a concurrent forfeit+consume stress test that **fails at −2250 against the pre-fix code** and passes with the lock+floor fix.

## Review
Money paths went through an adversarial review pass — it caught two blocking bugs (forfeit race + script under-forfeit), both fixed and regression-tested. Builds on the prior diagnosis + design in `docs/plans/credits-single-ledger-migration.md`.

## Notes
- Reconcile cron (missed-renewal/expiry backstop) is **deferred** to `louis/credits-reconcile-cron` — merge after, re-pointed at the new grant/forfeit helpers.
- Deploy: `PAYMENTS_GRANT_PLUS_MONTHLY` must be set — it boot-validates (process won't start without it).

---
<a href="https://app.blacksmith.sh/xmtplabs/codesmith/convos-backend/pr/324"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784803309&installation_id=128169237&pr_number=324&repository=xmtplabs%2Fconvos-backend&return_to=https%3A%2F%2Fgithub.com%2Fxmtplabs%2Fconvos-backend%2Fpull%2F324&signature=dac34235f4311bdbccc520d6230a2c405396b7f15b32b7a0b453598f2d2f9698"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

> [!NOTE]
> ### Implement single materialized wallet ledger for subscription credits grants and forfeits
> - Subscription credits are now written as real `sub_grant` ledger rows (keyed by `subscriptionId` + `periodStart`) on verify and renewal, replacing the previous approach of deriving balance from grant minus period consumes.
> - On expiry, revoke, or refund, `forfeitSubscriptionPeriod` claws back only unused subscription credits (clamped to wallet balance, never going negative), writing a `sub_forfeit` row idempotently.
> - `recordConsume` and `getSpendableBalance` in [spendable.ts](https://github.com/ephemeraHQ/convos-backend/pull/324/files#diff-cf5530a6d3c2c580e74ad2b31329241220f537979f3564f208593eb2f404abad) now unconditionally operate on the unified wallet balance; the `recordOnly` ledger write path is removed.
> - The `GET /v2/accounts/me/credits` handler now returns the actual wallet balance and derives `monthlyGrantUsed` from period consumes capped at the per-period grant.
> - A [database migration](https://github.com/ephemeraHQ/convos-backend/pull/324/files#diff-5a79ae9a1fe009746e8b56887d178cd7c5e232b0e773fb0015e65d6a9169fca0) inserts `sub_grant`/`sub_forfeit` grant kinds and widens the `CreditLedger_scope_check` constraint to include `sub_forfeit`.
> - Risk: `applyDeltaWithTx` no longer supports `recordOnly` writes; any callers that relied on this will now mutate the wallet balance.
>
> 
> #### Changes since #324 opened
>
> - Added staleness guard to `applyNotification` function in subscription repository that skips state updates and credit forfeitures when an incoming notification carries a `currentPeriodEnd` older than the stored subscription's `currentPeriodEnd`, while still recording the `BillingReceipt` [7675ce1]
> - Extended `upsertFromVerify` function in subscription repository to treat `BillingReceipt` idempotency key conflicts as benign concurrent duplicates, resolving them idempotently by re-reading the committed row instead of throwing an error [7675ce1]
> - Added `PAYMENTS_GRANT_PLUS_MONTHLY` environment variable to `.env.example` with default value of 2500 and updated comment to indicate five PAYMENTS_* configuration variables are required [7675ce1]
> - Added extensive inline documentation to `recordConsume` function in `src/payments/spendable.ts` describing the invariant that subscription period credits are materialized before consumption and documenting migration tradeoffs [7675ce1]
> - Added test coverage in `tests/subscriptions/repository.test.ts` for concurrent duplicate verify resolution, staleness guard behavior with out-of-order notifications, and legitimate mid-period revocation [7675ce1]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 4cb0b0e.</sup>
> 
>
> 

## Summary by CodeRabbit

* **New Features**
  * Subscription credits now use a single wallet balance, with per-period grants written directly into the ledger.
  * Credits responses keep the same JSON shape while showing a clearer spendable balance across free and subscribed accounts.

* **Bug Fixes**
  * Renewals, expirations, refunds, and revokes now apply credit changes more reliably and avoid double-counting.
  * Wallet balances are clamped safely to prevent negative spendable amounts.

* **Chores**
  * Updated boot-time validation for monthly subscription credit settings.


dev/scripts/materializeSubscriptionWallets.ts
```diff
@@ -0,0 +1,166 @@
+#!/usr/bin/env tsx
```
fbac: Assuming this was already run once, do we need this to persist on the repository?

prisma/migrations/20260623120000_credits_single_ledger/migration.sql
```diff
@@ -0,0 +1,29 @@
...
+ALTER TABLE "Subscription" ADD COLUMN "lastGrantedPeriodStart" TIMESTAMP(3);
+
+-- 2. New GrantKind rows for the subscription money-in / money-out on the one
+--    wallet. `sub_grant` is the per-period allotment; `subscription_forfeit`
+--    is the bounded clawback of the unused subscription portion on expiry/
+--    refund/revoke. ON CONFLICT keeps the migration idempotent.
+INSERT INTO "GrantKind" ("id", "name", "description", "active", "createdAt", "updatedAt")
+VALUES
+  ('sub_grant',            'Subscription grant',   'Per-period subscription credit allotment', true, CURRENT_TIMESTAMP, CURRENT_TIMESTAMP),
+  ('subscription_forfeit', 'Subscription forfeit', 'Bounded clawback of unused subscription credits on expiry/refund/revoke', true, CURRENT_TIMESTAMP, CURRENT_TIMESTAMP)
```
fbac: Unify grants naming so we avoid eroding the GrantKind over time. It can be sub_grant, sub_forfeit, or subscription_foobar.

prisma/migrations/20260623120000_credits_single_ledger/migration.sql
```diff
@@ -0,0 +1,29 @@
+-- Single-ledger wallet migration (n=1).
+--
+-- 1. Anchor column: the period for which we last wrote a `sub_grant` credit
+--    row. Lets the grant path (and the optional reconcile cron) decide
+--    "already granted this period?" without scanning the ledger.
+ALTER TABLE "Subscription" ADD COLUMN "lastGrantedPeriodStart" TIMESTAMP(3);
```
fbac: Is this absolutely necessary? Seems this value could be derived when needed with the current data we save.

src/api/v2/accounts/handlers/credits-get.ts
```diff
@@ -60,15 +63,13 @@ export async function creditsGetHandler(req: Request, res: Response) {
       requireSubscriptionTier(subscription.tier),
       subscription.period,
     );
-    const rawUsed = await sumPeriodConsumes(
-      accountId,
-      subscription.currentPeriodStart,
+    const monthlyGrantUsed = Math.min(
+      Math.max(0, grant.perPeriod - balanceCredits),
+      grant.perPeriod,
```
fbac, line 77: At this point we know the user has a valid subscription (is entitled) and that they were granted `grant.perPeriod` once in the previous < 30 days.

So the grant used can be derived as `Math.max(0, grant.perPeriod - balanceCredits)`
- 0 is the safeguard, as we don't want to go negative credits. Important followup: We need also code in place in the consume paths to not store negative units in database. AFAIK this has done but warrants a new visit.
- `grant.perPeriod - balanceCredits` is the source of truth. As we consume credits we'll be subtracting from the user balance, as mentioned before it must not go below 0.

I'm failing to see why we need the Math.min between the previously and `grant.perPeriod`. Can you give me context?

src/payments/spendable.ts
```diff
@@ -26,53 +21,33 @@ export const sumPeriodConsumes = async (
...
+ *
+ * Single-ledger migration: subscriptions write real `sub_grant` credit rows
+ * into `UserCredits.balance` on subscribe/renewal (and a bounded
+ * `subscription_forfeit` on expiry), so there is no longer a derived
+ * `tierGrant − periodConsumes` path or a bimodal switch on
+ * `isEntitledSubscription`. Subscribers and non-subscribers read the same
+ * wallet. Kept as a named export (rather than inlining `getBalance` at every
+ * call site) so the agent gate and admin view keep their stable import.
+ */
+export const getSpendableBalance = async (accountId: string): Promise<bigint> =>
```
fbac, line 35: Is this used anywhere else to make it exported?

src/subscriptions/grants.ts
```diff
@@ -0,0 +1,250 @@
...
+// Ledger idempotency keys are Stripe-style `[A-Za-z0-9_-]` (no colons), so the
+// plan's conceptual `sub_grant:{id}:{periodStartEpoch}` form is rendered with
+// underscores. Subscription ids are UUIDs (already dashed); the `{epoch}` suffix
+// keeps one row per (sub, period).
+
+/** Idempotency key for the per-period grant. One row per (sub, period). */
+export const subGrantKey = (
+  subscriptionId: string,
+  periodStart: Date,
+): string => `sub_grant_${subscriptionId}_${periodEpoch(periodStart)}`;
```
fbac, line 34: A gotcha here is that `periodStart` has to contain valid chars for the wallet idempotency keys zod validations, otherwise the regex will fail. We want to double check it.

This is the relevant code when creating idempotency keys:
```
export const IDEMPOTENCY_KEY_REGEX = /^[A-Za-z0-9_-]{1,255}$/;

export const idempotencyKeySchema = z.string().regex(IDEMPOTENCY_KEY_REGEX, {
  message: "invalid_idempotency_key",
});

export const assertIdempotencyKey = (key: string): void => {
  if (!IDEMPOTENCY_KEY_REGEX.test(key)) {
    throw new ValidationError("invalid_idempotency_key", {
      keyLength: key.length,
      disallowedChars: [...new Set(key.replace(/[A-Za-z0-9_-]/g, ""))],
    });
  }
};
```

src/subscriptions/grants.ts
```diff
@@ -0,0 +1,250 @@
...
+): string => `sub_forfeit_${subscriptionId}_${periodEpoch(periodStart)}`;
+
+const findLedgerRow = (
+  tx: TxClient,
+  accountId: string,
+  idempotencyKey: string,
+) =>
+  tx.creditLedger.findUnique({
+    where: { accountId_idempotencyKey: { accountId, idempotencyKey } },
+  });
```
fbac, line 49: Don't we already have public methods to access an idenmpotencyKey based on an accountId? If not we have to unify all these accesses under the same accessor.

src/subscriptions/grants.ts
```diff
@@ -0,0 +1,250 @@
...
+  const row = await findLedgerRow(tx, accountId, idempotencyKey);
+  if (!row) return 0;
+  const delta = row.delta;
+  return Number(delta < 0n ? -delta : delta);
+};
+
+export type GrantSubscriptionPeriodResult =
+  | { kind: "granted"; credits: number; subscription: Subscription }
+  | { kind: "replayed" }
+  | { kind: "skipped_nonpositive" };
```
fbac, line 99: Not savvy on TS but can't this type be handled as an enum? So we don't have to write an error prone `return { kind: "replayed" };` everywhere.

src/subscriptions/grants.ts
```diff
@@ -0,0 +1,250 @@
...
+    data: { lastGrantedPeriodStart: periodStart },
+  });
+
+  return { kind: "granted", credits, subscription: updated };
+};
+
+export type ForfeitSubscriptionPeriodResult =
+  | { kind: "forfeited"; credits: number }
+  | { kind: "replayed" }
+  | { kind: "skipped_nothing_to_forfeit" };
```
fbac, line 158: Same for this type.

src/api/v2/accounts/handlers/credits-get.ts
lourou: Display clamp for the "used X of Y" bar. A subscriber can spend past `perPeriod` (shared wallet, `minBalance ≤ 0`), so without `Math.min` it could render "used 1100 of 1000". Underflow is already prevented at the ledger write (`floorCheck`), not here.

src/payments/spendable.ts
lourou: Three real callers: admin account view, `credits-by-id`, and the `isSpendAllowed` spend gate (+ tests). It's a passthrough to `getBalance` for non-subscribers now, but the named export keeps one seam if spendability ever diverges from raw balance again. Keeping it.

dev/scripts/materializeSubscriptionWallets.ts
lourou: Removing it!

fbac (approved):

PR #355, opened 2026-07-13
fix(agent-templates): find the link anywhere in a build submission

saulmc:
The builder fetched a URL only when the **entire** submission was that URL (`looksLikeUrl` was an anchored `/^https?:\/\//` test). Every other way a person pastes a link routed wrong:

| Submission | Before |
|---|---|
| `Make me an agent about https://foo.com` | link never fetched — generates from the bare sentence |
| `https://foo.com check this out` | `new URL` throws → **build fails, 400 Invalid URL** |
| `https://foo.com/a?b=1 and more` | trailing words swallowed into the query string → fetches the wrong URL |

All three are ordinary human pastes, and the first is the same failure mode as the bare-`t.co` twitter-bot case: the source content never reaches the model, so the builder invents a generic agent.

`extractFirstUrl` now pulls the first parseable link from anywhere in the text and splits off the prose around it. The link is fetched as source material; the prose rides into the prompt as `User's intent:`, reusing the directive the attachment path already uses — so "make it answer support questions" survives instead of being discarded along with the rest of the sentence.

Two behaviors deliberately preserved:

- A paste longer than 300 chars that merely *cites* a link is still treated as the source material. The paste is what the user sent; fetching the citation would throw it away.
- A URL-shaped submission with no parseable link in it (`https://[invalid-url`) still fails with `Invalid URL` rather than quietly generating an agent from a malformed string.

---
<a href="https://app.blacksmith.sh/xmtplabs/codesmith/convos-backend/pr/355"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786502444&installation_id=128169237&pr_number=355&repository=xmtplabs%2Fconvos-backend&return_to=https%3A%2F%2Fgithub.com%2Fxmtplabs%2Fconvos-backend%2Fpull%2F355&signature=3975e49a5f96a0569b9b3c0eb73f5d3cb0f1c1a3397f2abb55bb9145f07b37f7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

> [!NOTE]
> ### Fix URL detection in `generateTemplate` to find links anywhere in a build submission
> - Adds `extractFirstUrl` to scan free-form text for the first valid URL token, strip trailing punctuation, and return the URL plus the surrounding residual prose.
> - Updates `generateTemplate` to fetch the extracted URL only when the residual prose is ≤300 chars; longer text is treated as pasted source material with no fetch.
> - Appends the residual prose as a `User's intent:` note in the prompt when a URL is fetched.
> - Returns a 400 error for URL-shaped text that contains no parseable URL token.
> - Behavioral Change: previously, URLs had to appear in a specific position in the submission; now any submission containing a URL token anywhere is handled correctly.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6b3061a.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Improved link detection in text-based template generation.
  * Preserves brief instructions surrounding a submitted link as user intent.
  * Fetches content from embedded links while retaining relevant instructions.
* **Bug Fixes**
  * Rejects malformed URL-like submissions with a clear “Invalid URL” error.
  * Correctly removes trailing punctuation from detected links.
  * Avoids fetching links when they are only incidental references in longer text.

### DIFF
--- a/src/api/v2/agent-templates/services/templateGen.ts
+++ b/src/api/v2/agent-templates/services/templateGen.ts
@@ -484,14 +484,25 @@ export interface SubmittedUrl {
   residual: string;
 }
 
-// A link stops at whitespace and at the delimiters that wrap one in prose
-// rather than belong to it. Trailing sentence punctuation is stripped
-// separately: `.` and `?` are legal in a URL but usually end the sentence.
+// A link stops at whitespace, and at the delimiters that wrap one in prose rather
+// than belong to it: angle brackets (`<https://…>`), quotes, backticks, and the
+// bracket pairs a link gets parenthesised or markdown-linked with. All of these
+// are legal inside a URL but overwhelmingly appear around one, and a link that
+// genuinely needs them can percent-encode them.
+//
+// Trailing sentence punctuation is stripped separately rather than excluded from
+// the token, because `.` `?` and `!` are common *inside* a URL and only ambiguous
+// at the very end ("see https://x.com/a." — the period ends the sentence).
 const URL_TOKEN_RE = /https?:\/\/[^\s<>"'`()[\]{}]+/i;
 const URL_TRAILING_PUNCT_RE = /[.,;:!?]+$/;
 
-/** First usable link anywhere in the text, with the surrounding prose split
- *  out. Null when the text holds no parseable link. Exported for tests. */
+/** First usable link anywhere in the text, with the surrounding prose split out.
+ *  Null when the text holds no parseable link.
+ *
+ *  Only the FIRST link is returned. A build is generated from one source, so a
+ *  submission carrying several links is read as being about the first one; any
+ *  others stay in `residual` and reach the model as part of the user's intent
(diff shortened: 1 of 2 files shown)


lourou (approved):
LGTM

PR #364, opened 2026-07-14
fix(agent-templates): featuring is the dashboard's, not the owner's

saulmc:
Follow-up to #363 — these two commits were pushed to that branch after it was squash-merged, so they didn't land with it. Nothing new since; this is the same work, cherry-picked onto the merged `otr-dev` and re-verified there.

Curation is the dashboard's. Two holes closed.

### The front door was open

Any signed-in user could put their own template into the convos.org featured gallery. `create` accepts `featured` in the body, and `PATCH` accepts it from whoever passes the ownership guard — which is the owner. Proven end to end with a plain JWT, no admin key anywhere:

| Plain user JWT | Before | After |
|---|---|---|
| `create` with `featured: true` | **201, featured stored** | **403** |
| publish own template | 200 | 200 (still theirs) |
| `PATCH featured: true` | **200** | **403** |
| `PATCH featuredRank: 999` | 403 | 403 |
| **Lands in the public gallery?** | **YES (rank 0)** | **NO** |

Only the *slot* was gated (#362), so a self-featured template entered at weight 0 — dead last. That contains it while the gallery holds a homepage's worth of curated templates above it, but containment is not a closed door: it inflates the featured set, and the moment the gallery holds fewer than eight weighted templates, an outsider renders on the homepage.

**Withdrawal stays the owner's.** `featured: false` can only ever take something *off* the homepage, and the runtime's builder sends it on every create — which is why the gate keys on `=== true`, not `!== undefined`. Gating that too would 403 every template the builder makes.

I checked that nothing legitimate features templates: across the runtime, website and workers, the only non-admin writer of `featured` is `build.mjs`, writing `false`. The admin console reaches the backend as an API-key caller and is unaffected — and the full suite passing with no test needing an update is itself evidence no existing flow relied on a user featuring anything.

### `PUT /featured-order` was mounted as a user endpoint

It admitted any signed-in caller at the middleware and then refused them inside with an `isApiKeyListener` check. That makes a pure curation tool part of the user-facing API surface and leaves its safety resting on a guard in the handler body.

It now takes the **agent key alone**: a signed-in caller is turned away at the door (401) rather than let in and told no, and the key is the identity, so there's no account to require. The in-handler check goes with it — once the route can't be reached without the key, a check for one is dead code shaped like a lock, and the next reader would reasonably infer the route was reachable without it.

The auth-wiring guard test pins the shape both ways, so it can't be quietly re-mounted behind the permissive middleware later:

```ts
expect(source).toContain("agentApiKeyAuth,\n  featuredOrderHandler");
expect(source).not.toContain("requireAccount,\n  featuredOrderHandler");
```

The dashboard is unaffected: the worker already injects the same `X-Agent-API-Key` (`AGENT_ASSETS_API_KEY`) that this stricter guard validates.

### Tests

- a user can't create a template that is already featured (403, and nothing is written)
- a user can't feature their own template afterwards (403, `featured` unchanged)
- a user may still create with `featured: false`, and unfeature their own
- the dashboard still features templates
- the order endpoint: a signed-in caller **and** an anonymous one both get 401, with the stored weights byte-identical afterwards

Full suite green (1492) on top of the merged `otr-dev`.

https://claude.ai/code/session_01DxFzZNgXaSnFWAYR8XPpCR

---
<a href="https://app.blacksmith.sh/xmtplabs/codesmith/convos-backend/pr/364"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786597085&installation_id=128169237&pr_number=364&repository=xmtplabs%2Fconvos-backend&return_to=https%3A%2F%2Fgithub.com%2Fxmtplabs%2Fconvos-backend%2Fpull%2F364&signature=3520f403a85c5bcbef73ae0a2c386c982b10a9e4a5a17bb30a61096d64c2925c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

> [!NOTE]
> ### Restrict agent template featuring to dashboard (agent API key) callers only
> - `PUT /featured-order` now uses `agentApiKeyAuth` exclusively; JWT-authenticated and anonymous callers receive 401.
> - `POST` and `PATCH` handlers for agent templates now return 403 if a non-agent-key caller attempts to set `featured: true`; setting `featured: false` remains permitted for owners.
> - The in-handler auth check is removed from [featured-order.ts](https://github.com/ephemeraHQ/convos-backend/pull/364/files#diff-11ed8f07178f6c010f5caf897115db3a4c2d79030baab96d32251cfd758b2b00) since the middleware now enforces it at the router level.
> - New test suite in [agent-templates.featuring-authz.test.ts](https://github.com/ephemeraHQ/convos-backend/pull/364/files#diff-864a56eb5f52eedf3274079cfb9e0f92f2445a14107964281dab1ff54a11d792) covers create/patch featuring authorization for both user and agent-key callers.
> - Behavioral Change: template owners can no longer feature their own templates via the public API; only the dashboard agent key can set `featured: true`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 699779c.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Restricted featured gallery ordering to authorized API-key access.
  * Added authorization controls for featuring agent templates during creation and updates.
  * Allowed authorized dashboard actions to feature or unfeature templates.

* **Bug Fixes**
  * Unauthorized attempts to feature templates now receive clear access-denied responses.
  * Unauthorized featured-order requests are rejected without changing gallery ordering.

### DIFF
--- a/src/api/v2/agent-templates/agent-templates.router.ts
+++ b/src/api/v2/agent-templates/agent-templates.router.ts
@@ -1,5 +1,6 @@
 import { Router } from "express";
 import {
+  agentApiKeyAuth,
   authOrAgentApiKeyAuth,
   optionalAuthOrAgentApiKeyAuth,
 } from "@/middleware/agentAuth";
@@ -77,12 +78,18 @@ agentTemplatesRouter.get(
   listCountsHandler,
 );
 
-// The featured gallery's order, written whole and in one transaction. Mounted
-// before /:id so the wildcard doesn't capture "featured-order" as a template id.
+// The featured gallery's order, written whole and in one transaction.
+//
+// Curation is not something a user does, so this isn't an endpoint a user
+// reaches: the agent key is required outright, and a signed-in caller is turned
+// away at the door rather than admitted and refused inside. The key IS the
+// identity here — there is no account to require.
+//
+// Mounted before /:id so the wildcard doesn't capture "featured-order" as a
+// template id.
 agentTemplatesRouter.put(
   "/featured-order",
-  authOrAgentApiKeyAuth,
-  requireAccount,
+  agentApiKeyAuth,
   featuredOrderHandler,
 );
 
--- a/src/api/v2/agent-templates/handlers/create.ts
+++ b/src/api/v2/agent-templates/handlers/create.ts
@@ -114,6 +114,25 @@ export async function createHandler(req: Request, res: Response) {
   //     (users can't create rows owned by someone else).
   //   - Anonymous: no account → 403 below.
   const isApiKeyListener = res.locals.isApiKeyListener ?? false;
+
+  // Featuring is curation, and the gallery is the homepage's — not the template
+  // owner's. Anyone may create a template; only the dashboard may put one in
+  // front of everybody. Without this an ordinary signed-in user could mint a
+  // self-featured template in a single request and publish it into the gallery
+  // convos.org renders.
+  //
+  // Asking for `featured: false` is a no-op and passes: the runtime's builder
+  // sends it on every create, and a template that isn't featured is exactly what
+  // a create would produce anyway.
+  if (parsed.data.featured === true && !isApiKeyListener) {
+    req.log.warn(
+      { callerAccountId: res.locals.accountId, action: "create.featured" },
+      "Unauthorized agent-template curation attempt",
+    );
+    res.status(403).json({ error: "Not authorized to feature a template" });
+    return;
+  }
+
   let ownerAccountId: string | undefined;
   if (isApiKeyListener && parsed.data.ownerAccountId !== undefined) {
     const assertedAccountId = parsed.data.ownerAccountId;
--- a/src/api/v2/agent-templates/handlers/featured-order.ts
+++ b/src/api/v2/agent-templates/handlers/featured-order.ts
@@ -56,22 +56,10 @@ export const isSerializationFailure = (error: unknown): boolean => {
  * won't match and the write is refused rather than clobbering a set the caller
  * never saw.
  */
+// Reachable only behind `agentApiKeyAuth` (see the router): curation is not a
+// thing a user does, so there is no caller here to authorize — a request without
+// the agent key never arrives.
 export async function featuredOrderHandler(req: Request, res: Response) {
-  // Curation is the dashboard's, not an owner's — same rule the rank field
-  // itself carries in PATCH.
-  const isApiKeyListener = res.locals.isApiKeyListener ?? false;
-  if (!isApiKeyListener) {
-    req.log.warn(
-      { callerAccountId: res.locals.accountId, action: "featured-order" },
-      "Unauthorized agent-template curation attempt",
-    );
-    sendError(res, 403, {
-      code: "FORBIDDEN",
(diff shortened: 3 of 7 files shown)


saulmc:
Took one of the three. Adjudicating all of them.

### ✅ Whitespace-brittle assertions — taken (699779c)

Fair, and worth more than it looks: a source-level assertion that fails on a *reflow* is noise, and a test that cries wolf is a test the next person weakens. They match the chain whitespace-agnostically now.

The interesting part is what fixing it surfaced. The naive regex is wrong:

```ts
const chain = (middleware: string, handler: string) =>
  new RegExp(`(?<![A-Za-z])${middleware},\\s+${handler}\\b`);
```

That lookbehind is load-bearing. Without it, **`agentApiKeyAuth` matches inside `authOrAgentApiKeyAuth`** — and those two differ by exactly the thing the assertion exists to tell apart: whether a signed-in caller can reach this endpoint. The original `toContain` happened to survive only because of a capital `A`. Both permissive middlewares are now named in explicit negative assertions rather than left to case-sensitivity to catch.

I mutation-checked the guard rather than trusting it: re-wiring `/featured-order` back to `authOrAgentApiKeyAuth + requireAccount` turns the test **red**; reverting turns it **green**.

### ❌ PATCH concurrent-modification 409 — the premise doesn't hold

> the `updateMany` pins to the row state including `featured` and `featuredRank`. If a concurrent request changes `featured`, this update gets a 409.

It doesn't. The `WHERE` pins `id`, `ownerAccountId`, `slug`, `status`, `firstPublishedAt` — **not** `featured`, and **not** `featuredRank`:

```ts
where: {
  id: template.id,
  ownerAccountId: template.ownerAccountId,
  slug: template.slug,
  status: template.status,
  firstPublishedAt: template.firstPublishedAt,
},
```

So the scenario described — an owner editing `description` while the dashboard features the template — does **not** 409. Only a concurrent change to status/slug/owner/first-publish trips it, which is the intent (those are the fields the handler's own transition rules were computed against). That optimistic-concurrency clause is also pre-existing (#199) and untouched here.

### ❌ Extract a shared `canFeature` helper — declining

Two call sites, and they aren't the same call: each carries its own log context (`create.featured` vs `patch.featured`) and its own refusal. Wrapping `featured === true && !isApiKeyListener` in a named predicate adds a hop without adding meaning — the rule is already stated once in each place it's enforced, next to the reason it's enforced there. Same reasoning I gave for `hasDuplicates` on #363.

Full suite green (1492).

